PR #11, opened 2016-09-17
Passing client to chansend and lazy version

Krukov:


andrewgodwin:
Why the lazy version?

Krukov:
Lazy version because when you use this package as source (by adding package path to the `PYTHONPATH`), importing this package got error with message ```pkg_resources.DistributionNotFound: The 'asgi_redis' distribution was not found and is required by the application``` (I think that is because module didn't install with pip) 
Full traceback
```python
>>> import asgi_redis
  File "<stdin>", line 1, in <module>
  File "asgi_redis/__init__.py", line 7, in <module>
    version = pkg_resources.require('asgi_redis')[0].version
  File "/home/frog/.virtualenvs/test2/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 968, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/frog/.virtualenvs/test2/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 854, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'asgi_redis' distribution was not found and is required by the application

```

andrewgodwin:
Ah, right. I'd suggest instead then that the version string is moved into the module and we import the module in setup.py like the other packages; it was only done like this before because it was a single module. I'd like to keep it as `__version__`.

Krukov:
Oh. I did it by by other way. Does  it matter?

andrewgodwin:
No, this seems fine, I've done the same before in other packages.

PR #13, opened 2016-09-27
DjangoRedisChannelLayer

furious-luke:
Hi again,

I'm not sure if these changes actually belong here, but I thought I'd make this PR to open up the discussion.

Essentially, I found that in order to have more control over how many active Redis connections there are, I needed to have a variation of the RedisChannelLayer class that uses the connection from `django-redis`. This way, when I limit the number of connections from `django-redis`, that limit is respected by `asgi_redis`.

Do you think this belongs here? Or perhaps even a standalone package that implements this? I'm not too sure...

Cheers,
Luke

### DIFF
--- a/asgi_redis/__init__.py
+++ b/asgi_redis/__init__.py
@@ -1,4 +1,4 @@
-from .core import RedisChannelLayer, DjangoRedisChannelLayer
+from .core import RedisChannelLayer
 from .local import RedisLocalChannelLayer
 
 __version__ = '0.14.1'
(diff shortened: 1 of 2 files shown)


andrewgodwin:
I would say this should be a separate package; I'd rather not introduce a dependency on another thing from inside asgi_redis, even if only optional, and it looks like it'd work fine as a pure subclass.

furious-luke:
Thanks for the feedback Andrew, I've changed the PR to remove the extra class, which I've placed in a separate package.

I've left in a very minor refactor which makes it much simpler to implement the changes in the other package.

andrewgodwin:
Thanks - I'll definitely pull in that change, I can see how it would make subclassing a lot easier.

PR #15, opened 2016-10-17
Handle Redis response errors instead of silencing them 

maikhoepfel:
The exception handling for send() silenced all Redis errors apart from the
"full" message. This commit updates the exception logic to also
specifically handle the "unknown command" error which might point at an
unsupported version of Redis. It also now lets any other exception bubble
up instead of being silently dropped.

Reported by @wynbennett in https://github.com/django/asgi_redis/issues/14.

I also added a wild guess as to what versions asgi_redis might support. I'm happy to also add Travis support for the whole grid of supported Django, Python and Travis versions if @andrewgodwin thinks it's a good idea.


andrewgodwin:
Thanks!

PR #24, opened 2017-01-13
Statistics

vartec:
Basic statistics endpoints implementation, with some caveats:

- does not provide calculated per second values (these are of little use if you're sending metrics to Graphite, CloudWatch or such)
- does not provide global `messages_pending` and `messages_max_age` due performance concerns, these are still available as per channel stats

### DIFF
--- a/asgi_redis/core.py
+++ b/asgi_redis/core.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 import base64
 import binascii
 import hashlib
-import math
 import msgpack
 import random
 import redis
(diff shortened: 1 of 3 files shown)


asgi_redis/core.py
```diff
@@ -317,6 +333,96 @@ def receive_twisted(self, channels):
...
+        Return value is a dictionary with following fields:
+            * messages_count, the number of messages processed since server start
+            * messages_pending, the current number of messages waiting
+            * messages_max_age, how long the oldest message has been waiting, in seconds
+            * channel_full_count, the number of times ChannelFull exception has been risen since server start
+
+        This implementation does not provide calculated per second values
+        """
+        statistics = {
+            'messages_count': 0,
```
bgreenberg-eb: Should these keys be constants?

PR #25, opened 2017-01-26
Test improvements

maikhoepfel:
This pull request follows https://github.com/django/asgiref/pull/13 in spirit: it adds tox, and configures Travis and tox so that tests also run against Python 3.3. 

I did find the Twisted tests are easy to miss. I hope to have improved upon the situation by extracting them into their own module, and raising a `SkipTest` so they show up during test collection. 

This is the first project where dependencies beyond `tox` are required to be able to run the test suite. I do think it makes sense to store those in a central place. As `requirements_dev.txt` was met with some resistance in #13, I opted to follow some popular Python projects and declared them in `setup.py`. This makes for a one-line project setup for newcomers, and ensures we don't need to update dependencies in more than one place.


setup.py
```diff
@@ -29,7 +31,8 @@ def get_version():
         'msgpack-python',
         'asgiref>=1.0.0',
     ],
-    extras_requires={
-        "cryptography": ['cryptography>=1.3.0'],
+    extras_require={
+        "cryptography": crypto_requires,
+        "tests": crypto_requires + ['tox', 'asgi_ipc', 'twisted', 'txredisapi'],
```
andrewgodwin: You can specify multiple extras using `asgi_regis[cryptography,tests]` - are you splitting them up like this for another reason?

andrewgodwin:
You're right about moving the Twisted tests into the right place, rest looks good apart from one small query on the extras.

setup.py
maikhoepfel: Yes, my idea was that `tests` should contain all dependencies needed to run the tests; it's simpler to the user that way. The tests happen to need `cryptography`, hence the overlap. I think we might need something similar for `daphne` and `channels`, and I'd like to standardize on being able to just install the `tests` extra for any of them to run the tests.

What do you think?

maikhoepfel:
I forgot to push changes to the README that I had locally. Sorry!

setup.py
andrewgodwin: I think that's fine for now!

PR #32, opened 2017-03-14
Add integration tests.

proofit404:
It is running http and websockets smoke test against testproject used in channels benchmark. This way we can test chat channels live server test case is implemented properly and useful for `asgi_redis` users.

**Do not merge this PR yet!**

It uses Channels and Daphne from my personal forks. I'll update it after https://github.com/django/daphne/pull/92 and https://github.com/django/channels/pull/497 will be merged (and maybe released).

Also consider to move test requirements into separate `requirements.txt` file. There is no way setuptools can install tarballs from the subdirectory of git repository. And we need this to install benchmark module and testproject.


proofit404:
Looks like setuptools test requires can resolve version conflict with preinstalled pytest on Travis-CI.

Can we move to the separate `requirements.txt` for test requirements? Pip doesn't have this issue.

andrewgodwin:
Well, pip is installing them, but it looks like you removed the `tests` extra from the setup.py script so it didn't actually do it right - add that back in and it should install it properly again.

proofit404:
I forgot to remove `[test]` extra from tox and Travis config. Pytest runner use test requires keyword instead. Extra requires named test is just a widely used convention. As far as I know.

proofit404:
Also there is no way to install benchmark package with setuptools​ until we move it to its own repository or submit to PyPI. So it's already kinda two files.

andrewgodwin:
Right, I'm saying drop `test_requires` and just use the extra. It's already in the travis script now with the `pip install -e .[tests]` command.

proofit404:
And `extras_require` doesn't respect `dependency_links` :'(

Sigh...

andrewgodwin:
Pip doesn't respect dependency links at all unless you turn on `--process-dependency-links` - they're considered deprecated.

proofit404:
Sorry, my bad. `dependency_links` are temporary solution here. I'll remove them before this PR will be merged.

proofit404:
I think we can finally merge this after django/channels#674 and manual test suite restart.

andrewgodwin:
I merged that in, so would a rebase trigger the build to pass?

proofit404:
Done.

andrewgodwin:
Hooray! All green.

PR #33, opened 2017-03-20
Add RedisSentinelChannelLayer

g-clef:
Adds a new Channel Layer: `RedisSentinelChannelLayer`.

A few notes on this: 
 * it only supports one sentinel cluster, so sharding is supported by having the sentinel cluster monitor multiple services, and specifying those services in the `CHANNEL_LAYERS` setting.
 * At least one service has to be specified in order to have something to ask the sentinel servers for.
 * The default redis sentinel object doesn't seem to support URL-format server specifications configuration, just tuple-format. So, the sentinel channel layer doesn't support URL-format either.
 * the "hosts" config setting in Sentinel format is the list of Sentinel servers, not redis servers.

### DIFF
--- a/asgi_redis/core.py
+++ b/asgi_redis/core.py
@@ -386,12 +386,15 @@ class RedisChannelLayer(BaseChannelLayer):
         these are only avaliable per channel.
 
         """
+        return self._count_global_stats(self._connection_list)
+
+    def _count_global_stats(self, connection_list):
         statistics = {
             self.STAT_MESSAGES_COUNT: 0,
             self.STAT_CHANNEL_FULL: 0,
         }
         prefix = self.stats_prefix + self.global_stats_key
-        for connection in self._connection_list:
+        for connection in connection_list:
             messages_count, channel_full_count = connection.mget(
                 ':'.join((prefix, self.STAT_MESSAGES_COUNT)),
                 ':'.join((prefix, self.STAT_CHANNEL_FULL)),
@@ -412,6 +415,14 @@ class RedisChannelLayer(BaseChannelLayer):
 
         This implementation does not provide calculated per second values
         """
+        if "!" in channel or "?" in channel:
+            connections = [self.connection(self.consistent_hash(channel))]
+        else:
+            # if we don't know where it is, we have to check in all shards
+            connections = self._connection_list
+        return self._count_channel_stats(channel, connections)
+
+    def _count_channel_stats(self, channel, connections):
         statistics = {
             self.STAT_MESSAGES_COUNT: 0,
             self.STAT_MESSAGES_PENDING: 0,
@@ -420,14 +431,7 @@ class RedisChannelLayer(BaseChannelLayer):
         }
         prefix = self.stats_prefix + channel
 
-        if "!" in channel or "?" in channel:
-            connections = [self.connection(self.consistent_hash(channel))]
-        else:
-            # if we don't know where it is, we have to check in all shards
-            connections = self._connection_list
-
         channel_key = self.prefix + channel
-
         for connection in connections:
             messages_count, channel_full_count = connection.mget(
                 ':'.join((prefix, self.STAT_MESSAGES_COUNT)),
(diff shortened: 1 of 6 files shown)


asgi_redis/sentinel.py
```diff
@@ -0,0 +1,208 @@
...
+            )
+            # Decide on a unique client prefix to use in ! sections
+            # TODO: ensure uniqueness better, e.g. Redis keys with SETNX
+            self.client_prefix = "".join(random.choice(string.ascii_letters) for i in range(8))
+            # Register scripts
+            connection = self.connection(None)
+            self.chansend = connection.register_script(self.lua_chansend)
+            self.lpopmany = connection.register_script(self.lua_lpopmany)
+            self.delprefix = connection.register_script(self.lua_delprefix)
+            self.incrstatcounters = connection.register_script(self.lua_incrstatcounters)
```
andrewgodwin: Could you refactor this and the base class to bring all the common stuff like this into one place so it's not duplicated?

andrewgodwin:
Seems good overall, but I'd like to know how hard it would be to refactor away some of the duplicated code (specifically the __init__ code, but also maybe the stats stuff); feel free to modify the base class if you need to. Otherwise, I can look into that.

You will have to get Travis working to get it merged, though; I suspect you probably need to have it skip the tests if it detects that there is no Sentinel available to test on, or find a way to run a sentinel instance using the `.travis.yml` configuration.

g-clef:
The duplicate code refactor should be easy enough. I'll check on the travis stuff.

g-clef:
From what I can see of the Travis documentation, it's possible to run redis-server, but not redis-sentinel. So, I modified the sentinel unit tests to check that it can retrieve a Redis master from sentinel for the service_name used in the tests, and if it can't to skip those tests.

asgi_redis/core.py
```diff
@@ -93,12 +81,35 @@ def __init__(
         # Decide on a unique client prefix to use in ! sections
         # TODO: ensure uniqueness better, e.g. Redis keys with SETNX
         self.client_prefix = "".join(random.choice(string.ascii_letters) for i in range(8))
-        # Register scripts
+        self._register_scripts()
+        self._setup_enrcyption(symmetric_encryption_keys)
```
andrewgodwin: encryption is misspelt here in `_setup_enrcyption`

asgi_redis/core.py
```diff
@@ -93,12 +81,35 @@ def __init__(
...
+        return final_hosts
+
+    def _register_scripts(self):
         connection = self.connection(None)
         self.chansend = connection.register_script(self.lua_chansend)
         self.lpopmany = connection.register_script(self.lua_lpopmany)
         self.delprefix = connection.register_script(self.lua_delprefix)
         self.incrstatcounters = connection.register_script(self.lua_incrstatcounters)
+
+    def _setup_enrcyption(self, symmetric_encryption_keys):
```
andrewgodwin: Also misspelt here.

asgi_redis/sentinel.py
```diff
@@ -0,0 +1,142 @@
...
+                                                            prefix,
+                                                            group_expiry,
+                                                            capacity,
+                                                            channel_capacity,
+                                                            symmetric_encryption_keys,
+                                                            stats_prefix,
+                                                            socket_connect_timeout,
+                                                            socket_timeout,
+                                                            socket_keepalive,
+                                                            socket_keepalive_options)
```
andrewgodwin: Could you indent these to match the existing code style (that is, just bring them in 4 spaces compared to the outer scope, rather than aligning them all with the opening parenthesis?)

g-clef:
Heh. That's embarrassing. I made those changes. It looks like the travis builds are failing for another reason, so I'm not going to try to fix that myself.

andrewgodwin:
Yes, the builds are failing right now due to the changes on the `asgiref` master that aren't landed, and which I can't release yet. I'll pull this down, test and merge myself.

andrewgodwin:
Alright, manually merged. Thanks!

PR #42, opened 2017-05-17
Sentinel bug

g-clef:
If you have multiple hosts, but a single service in redis sentinel, ring_size gets set to len(hosts) not len(services), leading to `list index out of range` exceptions. Moved sentinel's configuration of ring_size to run after the super __init__ to reset this to the proper value.

### DIFF
--- a/asgi_redis/sentinel.py
+++ b/asgi_redis/sentinel.py
@@ -37,9 +37,6 @@ class RedisSentinelChannelLayer(RedisChannelLayer):
     ):
         self.services = self._setup_services(services)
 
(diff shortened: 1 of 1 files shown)


andrewgodwin:
I believe this is also covered by the bigger rewrite in #43, which overhauls a lot of the sentinel stuff - if you agree, we can close this since I'm going to merge that.

g-clef:
looks like it. The issue was needing to run the super().__init__ before initializing ring_size, and it looks like that pull does that.

andrewgodwin:
OK, I'll close this then.

PR #43, opened 2017-05-17
Add service discovery and connection caching to Sentinel, and make send/receive more deterministic

bgreenberg-eb:
There are several large changes here. The primary purpose is to add some key capabilities to the Sentinel channel layer. In the process, I've made some improvements to the core Redis channel layer and added a Docker-based testing environment.

- Added a Docker-based testing environment using `docker-compose`. Provides two Redis masters, two slaves, and one Sentinel server. The Redis servers are configured to not persist data to disk.
- Made send/receive logic more deterministic for normal (non-single-reader) channels. When sending to or receiving from a normal channel, the channel layer now cycles through all connections in order, rather than choosing at random. This guarantees that a single worker backed by multiple connections will pick up a message after calling `receive()` no more than once for each connection. This makes the channel layer behavior more predictable without compromising performance, and makes it possible to test channel layers with multiple connections.
- Added two new capabilities to the Sentinel channel layer: connection caching and service discovery. 
  - Connection caching is enabled by passing `sentinel_refresh_interval`. This will cause the channel layer to cache its master connections and only refresh them when a connection is requested and the refresh interval has passed since the last refresh.
  - Service discovery happens automatically if `services` is empty. The channel layer calls out to its Sentinel hosts to get all service names tracked by Sentinel, and uses that information to build its service list.


bgreenberg-eb:
Looks like I need to make some changes to the tests that they work both in Travis and in Docker.

PR #45, opened 2017-05-24
Proposed fix for #44

### LINKED ISSUE #44: RedisChannelLayer.receive runs forever
`RedisChannelLayer.receive`, that was reworked in #43, now hangs forever if called with block. That was not the case in the previous version (it returned `(None, None)`. This prevents channels `Worker` from exiting if termed is set to true (e.g. in the signal handler).

AlexejStukov:
This should fix #44. Hope it does not break anything else.

### DIFF
--- a/asgi_redis/core.py
+++ b/asgi_redis/core.py
@@ -228,7 +228,7 @@ class RedisChannelLayer(BaseChannelLayer):
                         del message['__asgi_channel__']
                     return channel, message
             # If we only got expired content, try again
-            if block or got_expired_content:
+            if got_expired_content:
                 continue
             else:
                 return None, None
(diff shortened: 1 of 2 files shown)


proofit404:
Hi,

Thanks for your contribution. Tests to prove that problem was present and solved wold be greet.

AlexejStukov:
Travis build hangs as expected

AlexejStukov:
This one should complete.

AlexejStukov:
Is it OK now, @proofit404?

andrewgodwin:
Rejecting this - see my response to #44.

proofit404:
Actually, the whole point of `block=True` argument is to wait until the message is available.

PR #59, opened 2017-09-13
Factor out common functionality from Redis and Sentinel channel layers into base class

bgreenberg-eb:
### Summary

Since `RedisSentinelChannelLayer` performs a fundamentally different setup routine than `RedisChannelLayer` does, it does not make sense for the former to inherit from the latter. The previous implementation was clunky and limiting, so I moved the common methods and setup steps into `BaseRedisChannelLayer`, and made `RedisSentinelChannelLayer` and `RedisChannelLayer` inherit from it.

### Bug fixes
- The original motivation for this work was to fix a bug introduced by [this commit](https://github.com/andymccurdy/redis-py/commit/29a7ecebd0f4709fc184c6af57c6eb1a59cb071e#diff-085cbe85ef9bc2ad832d79163eb39172) in `redis-py`, which was included in the 2.10.6 patch version. I updated the channel layer script initialization to work with this version of Redis. This required moving script initialization to the end of the `__init__` method of `RedisSentinelChannelLayer`, which prompted the refactor in the first place.
-  [This commit](https://github.com/django/asgi_redis/commit/984b5be51e7aba6144f9e99bcdc90139c10454d4) introduced a `git` requirement for tests, so I added that to the `Dockerfile`. The integration tests included in that commit also break in Docker, because they rely on settings from another repository that point to the default address for Redis. I made sure those tests are skipped in Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@ FROM ubuntu:16.04
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
     apt-get -yqq install \
+    git \
     build-essential python-pip software-properties-common \
     python-dev python3-dev \
     libffi-dev libxml2-dev libxslt-dev libssl-dev
--- a/asgi_redis/core.py
+++ b/asgi_redis/core.py
@@ -7,7 +7,6 @@ import itertools
 import msgpack
 import random
 import redis
-from redis._compat import b
 import six
 import string
 import time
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM ubuntu:16.04
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
     apt-get -yqq install \
+    git \
```
vartec: ?

bgreenberg-eb: Apparently git isn't installed by default in that image. Tests were failing in Docker because the command was not found.

PR #61, opened 2017-10-17
prevent error if oldest message is no longer present

### LINKED ISSUE #60: ttl could be `None`
I got this report on my error logger
`TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'`

on this line:
https://github.com/django/asgi_redis/blob/master/asgi_redis/core.py#L395

so it appears that `ttl` can return `None`

freakypie:
I am not a clever man, but attempting to fix #60


andrewgodwin:
Seems good to me! Thanks.

Issue #71, opened 2018-02-10
Connection <RedisConnection [db:0]> has pending commands, closing it.


bastbnl:
While trying to understand what's happening on channels issue 858 I found this message: 
```
[2018-02-09 21:21:04,819: WARNING/ForkPoolWorker-2] Connection <RedisConnection [db:0]> has pending commands, closing it.
```

I scanned through the getting started guide at https://github.com/jonathanslenders/asyncio-redis and found that each connection was closed, in the examples. I did the same thing on my install and noticed that the message was gone, while it still appeared to work.

andrewgodwin:
We're doing things a little differently to the examples though because there's a connection pool spread across threads. I'll see if I get this message today while debugging django/channels#858.

PR #72, opened 2018-02-10
Explicitly closes the connection (#71)

bastbnl:


andrewgodwin:
I'm not sure why you would want to do this? The underlying implementation uses a connection pool and a transaction manager so that it holds open connections for re-use.

bastbnl:
Me neither. I'm also wondering if it's used as a way to tell the API that we're done with this reference for now, while still keeping the connection alive. It did make the message go away, but I'm not sure if it's wrecking something else.

It would not make sense to actually close the connection as the connection knows it's from a pool. Right?

andrewgodwin:
Right. I'm going to close this out for now as I'm going to do a deep dive into the connection/async code over the weekend, and hopefully I can figure out the right way to manage all the connections. Thanks though.

PR #85, opened 2018-03-13
Group Send using LUA

mmcardle:
* Change the group_send api function to use a LUA script

- [x] Add a test which send to many group (1000+) and check it returns within a timeout


channels_redis/core.py
```diff
@@ -310,22 +310,28 @@ def check_receive_lock(self):
...
-                x.decode("utf8") for x in
-                await connection.zrange(key, 0, -1)
-            ]
-        # TODO: More efficient implementation (lua script per shard?)
-        for channel in channel_names:
-            try:
-                await self.send(channel, message)
-            except ChannelFull:
-                pass
+            await connection.eval(
```
andrewgodwin: You're only sending this on a single connection, so it won't work with a multiple-shard setup (the old code's use of individual `send` calls means it picks the right shard for each destination channel).

As the comment you deleted says, a Lua script would have to be run on each shard after getting all the destination channel names and grouping them by which shard their consistent hash puts them on.

mmcardle: righty! ill get that addressed. Do you require tests that test it working over multiple connections?

andrewgodwin: That would be ideal, though getting multiple-Redis tests working in Travis has proved hard in the past (which is why there aren't any right now). You might be able to do it with different database numbers on the single redis host they provide.

mmcardle: I will try with different database numbers

mmcardle:
@andrewgodwin I have updated with tests, i'm not sure I understand why I needed to make the change in the `receive_loop` function (apart from the fact it needed the changes to function)

channels_redis/core.py
```diff
@@ -81,9 +82,15 @@ def decode_hosts(self, hosts):
         # Decode each hosts entry into a kwargs dict
         result = []
         for entry in hosts:
-            result.append({
-                "address": entry,
-            })
+            if len(entry) == 3:
```
andrewgodwin: You don't need this alternative format - use Redis URIs instead, like `redis://host:6379/0`

channels_redis/core.py
```diff
@@ -154,12 +161,21 @@ def _setup_encryption(self, symmetric_encryption_keys):
                 # Wait on the receive buffer's contents
                 # TODO: Two coroutines rather than a poll
                 while True:
-                    if self.receive_buffer.get(channel, None):
-                        message = self.receive_buffer[channel][0]
-                        if len(self.receive_buffer[channel]) == 1:
-                            del self.receive_buffer[channel]
+
+                    messages = self.receive_buffer.get(channel, None)
```
andrewgodwin: Why did you change this section? It looks like you're working around the single-process-channel implementation

channels_redis/core.py
mmcardle: Yup will do, was following the format of `TEST_HOSTS` but redis urls make  more sense

channels_redis/core.py
mmcardle: Yes this is part I was trying to figure out, without these changes the `test_groups_multiple_hosts*` tests fail as the receive_buffer never get populated with the message, and the `(await channel_layer.receive(channel))` times out after a group_send

mmcardle:
Going to rebase this to the latest 2.1.1

mmcardle:
@andrewgodwin rebased to the latest 2.1.1 , the new code now doesn't touch anything apart from the group_send function.

channels_redis/core.py
```diff
@@ -316,12 +317,59 @@ def _setup_encryption(self, symmetric_encryption_keys):
                 x.decode("utf8") for x in
                 await connection.zrange(key, 0, -1)
             ]
-        # TODO: More efficient implementation (lua script per shard?)
+
+        connection_to_channels, channel_to_message = (
+            self._map_channel_to_connection(channel_names, message)
+        )
```
andrewgodwin: This is less than 120 characters, no need to wrap it

channels_redis/core.py
```diff
@@ -316,12 +317,59 @@ def _setup_encryption(self, symmetric_encryption_keys):
...
+                # Set the message expiry
+                group_send_lua += (
+                    b"redis.call('EXPIRE', '%s', %d)\n"
+                    % (channel_name.encode("utf8"), self.expiry)
+                )
+
+            async with self.connection(connection_index) as connection:
+                await connection.eval(
+                    group_send_lua, keys=channel_names
+                )
```
andrewgodwin: You're not using a pre-loaded Lua script here but rather just making one dynamically - have you considered using a pipeline instead? Have you done any basic speed tests between both?

channels_redis/core.py
mmcardle: will do

mmcardle:
Also just noticed there are no checks for channel capacity, which the previous use of `send` would have.

channels_redis/core.py
mmcardle: I have done some benchmarking against a single host against 3 group send methods
basic send, LUA and Pipeline in this gist https://gist.github.com/mmcardle/4e59943275cc1ad221b7bdade21cb45e

andrewgodwin: Yup, I am happy that lua is a lot better!

mmcardle: Thanks @andrewgodwin I am just trying to debug an issue with the LUA version of group_send, getting a lovely `Segmentation fault` when using it in my application, works fine with pipeline.
Seeing if i can reproduce it.

proofit404: Is there really no way to assign lua script content statically in redis and call it with arguments?

mmcardle: Have debugged the `Segmentation fault` was to do with hiredis parsing the results and throwing a unicode decode error. Have updated the group_send to pass in the messages as arguments to the `connection.eval(...)` function and everything seems alot happier

mmcardle:
Updated

agronick:
Looking forward to this

andrewgodwin:
Merged - thanks for all your work on this.

PR #90, opened 2018-03-28
Fixed the receive_buffer to be a defaultdict using queues instead of lists

### LINKED ISSUE #89: Messages sent though Redis sometimes come out of order
With Channels 2 I am noticing an issue that wasn't present in Channels 1. We use channels to display job status. We will start hundreds of job and display the status in a progress bar. Sometimes messages will come in out of order causing the progress bar to move backwards. This happens when many jobs are finishing at the same time. When the same code is used with a consumer that does not use broadcasting the problem goes away.

The problem is not on the client side. In the function called by the "type" key I can see the messages coming in out of order.

agronick:
This fixes #89. Messages no longer come in out of order. Seems to be somewhat faster as well. Uses queues instead of polling. Still checks for bubbling exception. `local_poll_interval` is used as a timeout instead of a sleep.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -6,6 +6,7 @@ import itertools
 import random
 import string
 import time
+import collections
 
 import aioredis
 import msgpack
@@ -68,7 +69,7 @@ class RedisChannelLayer(BaseChannelLayer):
         # Main receive loop running
         self.receive_loop_task = None
         # Buffered messages by process-local channel name
-        self.receive_buffer = {}
+        self.receive_buffer = collections.defaultdict(asyncio.Queue)
 
     def decode_hosts(self, hosts):
         """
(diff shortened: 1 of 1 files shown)


channels_redis/core.py
```diff
@@ -169,22 +170,14 @@ def _setup_encryption(self, symmetric_encryption_keys):
...
-                        # There's a message, pop it off and shorten the buffer
-                        message = self.receive_buffer[channel][0]
-                        if len(self.receive_buffer[channel]) == 1:
-                            del self.receive_buffer[channel]
-                        else:
-                            self.receive_buffer[channel] = self.receive_buffer[channel][1:]
-                        return message
-                    else:
+                    try:
+                        return await asyncio.wait_for(self.receive_buffer[channel].get(), self.local_poll_interval)
```
andrewgodwin: You need to clean up the entries in the dict if the queue is empty, or this will memory leak over time.

channels_redis/core.py
```diff
@@ -28,7 +29,7 @@ class RedisChannelLayer(BaseChannelLayer):
     """
 
     blpop_timeout = 5
-    local_poll_interval = 0.01
+    local_poll_interval = 10
```
andrewgodwin: If you are changing the meaning of the variable, rename it!

agronick:
Will do

agronick:
Not sure why it doesn't like the imports. But I put a few hundred thousand messages through this and it works much better.

agronick:
Any idea why it doesn't like tge imports?

andrewgodwin:
There's a diff in the Travis output saying what it wants you to do - it's detecting too many blank lines it looks like.

andrewgodwin:
Thanks!

PR #91, opened 2018-04-03
Reword the group_expiry explanation

matthiask:
Mention the websocket timeout in daphne. The README recommends lowering
``group_expiry`` but does not mention that websocket timeouts should
also be lowered at the same time. I also had active websocket connections
which did not receive any group message, and had to read the code and
monitor redis to find out the reason for this.

From: https://github.com/django/channels/issues/999#issuecomment-378094611

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,9 +57,12 @@ get to it.
 ``group_expiry``
 ~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Lovely, thanks.

Issue #100, opened 2018-04-15
It reconnects a lot of time in `group_send`


imbolc:
Each call of `connection` context processor cause reconnection to redis. So in `group_send` there are number_of_channels + 1 reconnections (with the same parameters in my case). Why is it so and is there a way to optimize this?

imbolc:
May be related to #83

imbolc:
Just checked it works 7 times faster if I pass connection from `group_send` to `send`

andrewgodwin:
Is this on the released version of `channels_redis`? The git master branch uses a Lua-based sending script and should be significantly faster and reconnect less.

imbolc:
Yes, it was metrics from the pip version, on git one are 2 connection per message, and if I make it one the speed is increasing from 7.34 ms to 5.3 ms per message. But why it connects for each message, wouldn't it better to just keep first connection?

andrewgodwin:
Ah yes, the git version is efficient at sending to lots of clients but will still connect every time you call `group_send` because there's no connection persistence between calls. It used to be in there but it ended up causing connection leaks and making unit testing harder, and so I removed the persistence in favour of making it slow and correct, with the hope we'll get time to properly test and put persistence back in later.

I'll change this issue to represent the fact we should have connection persistence, but I wouldn't expect to get it worked on in the next few months unless someone other than me has time, as I have a lot of other stuff to do as well.

Anisa:
Experienced the same problem. Added a PR which fixes the issue for me.

banksee:
I think it should be `channels`' responsibility to call a `self.channel_layer.close()`, so, having that - what needs to be done in @Anisa's PR is adding `.close()` calls to the layer fixtures

andrewgodwin:
The `close()` method is only really needed for tests, and even then I think I removed the need for it there. It's not needed in normal use.

banksee:
@andrewgodwin how is connections are closed in a "normal use"? which part of the system does that?

if having `.close()` in the tests for @Anisa's PR is enough - I will be able to submit my changes.

andrewgodwin:
@banksee The Redis connections are closed once the `receive()` or `send()` call returns, normally. It's done via a context manager if remember right. I don't believe it will solve the problems in that PR, but you're welcome to try!

banksee:
@andrewgodwin yes, it is the case with the current implementation, but I wonder how we should close *persistent* connections, I am thinking of some counter for this, i.e.:
[...]

PR #104, opened 2018-05-08
Added persistence on connections.

Anisa:
Created a new persistent connection class so that the connections
are not recreated every time a new request is sent.
This new persistent connection will check if it's already created
and that the connection answers to PING, if it does; reuse it. Else
recreate it.

Fixes: #100

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -10,6 +10,7 @@ import time
 
 import aioredis
 import msgpack
+from aioredis.errors import ConnectionClosedError
 
 from channels.exceptions import ChannelFull
 from channels.layers import BaseChannelLayer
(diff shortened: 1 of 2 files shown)


andrewgodwin:
I can't take this as-is because it raises loads of warnings during tests: https://travis-ci.org/django/channels_redis/jobs/376310386

The reason I removed persistence in the first place was because it's very difficult to make sure things persist *and* you clean up dangling threads. If you can get this to work without warnings from the tests, then I'll be happy to merge it!

Anisa:
I do not have any time to finish this up, sadly.
If anyone wants to go ahead an try finishing it after @andrewgodwin comments, go ahead.

Issue #7, opened 2016-09-09
Make group sending more efficient using Lua scripts


andrewgodwin:
Right now it does individual sends which is shorter code wise but not as efficient. This is more complicated than it seems as it must be done per shard.

gekco:
I have made some changes regarding this. Have a look

Issue #83, opened 2018-03-01
Sending messages is slow when using groups


remusmp:
It seems that `groups` slow down sending messages ~100 times:
I've created 3 different echo consumers in order to compare them:
1. AsyncJsonWebsocketConsumer with groups
2. AsyncConsumer with groups
3. AsyncConsumer without groups

I got these timing results:

```
1.
Elapsed time 0.026857614517211914.
Elapsed time 0.026637554168701172.
Elapsed time 0.03318667411804199.
Elapsed time 0.03062891960144043.
Elapsed time 0.026413679122924805.
Elapsed time 0.028857707977294922.
Elapsed time 0.03323197364807129.
Elapsed time 0.026538610458374023.
Elapsed time 0.036324501037597656.
Elapsed time 0.0322566032409668.
Elapsed time 0.0385587215423584.
Elapsed time 0.02066636085510254.
Elapsed time 0.03473186492919922.
Elapsed time 0.03095531463623047.
Elapsed time 0.022891759872436523.

2.
Elapsed time 0.02672290802001953.
Elapsed time 0.022529125213623047.
Elapsed time 0.042157888412475586.
Elapsed time 0.028131961822509766.
Elapsed time 0.035176753997802734.
Elapsed time 0.03284716606140137.
Elapsed time 0.03023362159729004.
Elapsed time 0.03440999984741211.
Elapsed time 0.029916763305664062.
Elapsed time 0.03461766242980957.

3. 
Elapsed time 0.00019168853759765625.
Elapsed time 9.107589721679688e-05.
Elapsed time 0.00020265579223632812.
Elapsed time 0.0004429817199707031.
Elapsed time 0.0002741813659667969.
Elapsed time 0.00019693374633789062.
Elapsed time 0.00026679039001464844.
Elapsed time 0.0003097057342529297.
Elapsed time 0.00017261505126953125.
Elapsed time 0.0003085136413574219.
Elapsed time 0.00048232078552246094.
Elapsed time 0.0002384185791015625.
```

Here is the consumers code:

```
from channels.generic.websocket import AsyncJsonWebsocketConsumer, AsyncConsumer
from time import time
from channels.layers import get_channel_layer
from asgiref.sync import async_to_sync, sync_to_async

class EchoConsumer1(AsyncJsonWebsocketConsumer):

    async def connect(self):
        await self.accept()
        await self.channel_layer.group_add("test", self.channel_name)

    async def disconnect(self, close_code):
[...]

PR #107, opened 2018-06-05
Sending Single Message Per Client Prefix/ User Prefix

gekco:
previously for each channel subscribed to a group message was  being pushed  to list stored at channel key. I have grouped those messages by sending list of channels associated with that that channel redis key with a single message  . It saves a great deal of time by reducing the number of push and pops from redis.
Fixes : #7, #83

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -382,9 +382,8 @@ class RedisChannelLayer(BaseChannelLayer):
 
         1. list of their redis keys bucket each one to a dict keyed by the connection index
 
-        2. for each unique channel redis key create a message specific to that redis key, by adding
-           the comma separated string of channels which maps to that particular redis key
-           in __asgi_channel__ key to the message
+        2. for each unique channel redis key create a serialized message specific to that redis key, by adding
+           the list of channels mapped to that redis key in __asgi_channel__ key to the message
 
         3. returns a mapping of redis channels keys to their capacity
         """
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -200,7 +200,11 @@ class RedisChannelLayer(BaseChannelLayer):
         assert general_channel.endswith("!"), "receive_loop not called on general queue of process-local channel"
         while True:
             real_channel, message = await self.receive_single(general_channel)
-            await self.receive_buffer[real_channel].put(message)
+            if type(real_channel) is list:
+                for channel in real_channel:
+                    await self.receive_buffer[channel].put(message)
+            else:
+                await self.receive_buffer[real_channel].put(message)
 
     async def receive_single(self, channel):
         """
@@ -218,6 +222,7 @@ class RedisChannelLayer(BaseChannelLayer):
         async with self.connection(index) as connection:
             channel_key = self.prefix + channel
             content = None
+
             while content is None:
                 content = await connection.blpop(channel_key, timeout=self.blpop_timeout)
             # Message decode
@@ -225,7 +230,7 @@ class RedisChannelLayer(BaseChannelLayer):
             # TODO: message expiry?
             # If there is a full channel name stored in the message, unpack it.
             if "__asgi_channel__" in message:
-                channel = message["__asgi_channel__"]
+                channel = message["__asgi_channel__"].split(",")
                 del message["__asgi_channel__"]
             return channel, message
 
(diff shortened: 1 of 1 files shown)


channels_redis/core.py
```diff
@@ -310,33 +315,34 @@ def _setup_encryption(self, symmetric_encryption_keys):
...
-                    end
-                end
-            """ % self.expiry
+                        for i=1,#KEYS do
+                            if redis.call('LLEN', KEYS[i]) < tonumber(ARGV[i + #KEYS]) then
+                                redis.call('RPUSH', KEYS[i], ARGV[i])
+                                redis.call('EXPIRE', KEYS[i], %d)
+                            end
+                        end
+                    """ % self.expiry
```
andrewgodwin: Why did you indent this without changing it?

channels_redis/core.py
```diff
@@ -350,26 +356,38 @@ def _map_channel_to_connection(self, channel_names, message):
...
+                channel_key_to_message[channel_key] = message
+                channel_key_to_capacity[channel_key] = self.get_capacity(channel)
+            else:
+                channel_key_to_message[channel_key]["__asgi_channel__"].append(channel)
+
+        for key in channel_keys:
+            channel_key_to_message[key]["__asgi_channel__"] = ",".join(channel_key_to_message[key]["__asgi_channel__"])
+            channel_key_to_message[key] = self.serialize(channel_key_to_message[key])
+
+        return channel_keys, connection_to_channel_keys, channel_key_to_message, channel_key_to_capacity
```
andrewgodwin: This block needs some serious comments to say what's happening - I'm finding it hard to read.

channels_redis/core.py
gekco: i guess the IDE did it. i will revert it.

channels_redis/core.py
gekco: on it

gekco:
(Correct me if i am understanding this wrong)
say there is a group name 'Group-A' with 100 members(channels)

originally when i called group_send('Group-A', message_dictionary) 
the same message was being pushed 100 times each time with a different __asgi_channel__ value   to the same  redis queue(because multiple channels have same channel_key ) which was being popped at function receive_single 100 times obviously and the message was getting processed for a single channel.

Modifications:
what i did was coalesce those messages that is for those 100 channels  i will be sending a single message for all the channels with the same channel key by sending a comma separated string of channels instead of a single channel. On the receiving end i just process the message for all of the channels that are received in the message. This Largely reduces the time to push group messages onto channel redis layer and the number of pops. 
For Example if all the 100 channels have the same channel key there will be a single message for all the 100 channels which will be pushed on the queue and popped from the queue.

channels_redis/core.py
```diff
@@ -310,66 +316,102 @@ def _setup_encryption(self, symmetric_encryption_keys):
...
 
             # We need to send the capacity for each channel
-            args += [channel_to_capacity[channel_name] for channel_name in channel_names
-                    if channel_to_key[channel_name] in channel_redis_keys]
+            args += [channel_keys_to_capacity[channel_key] for channel_key in channel_keys
+                    if channel_key in channel_redis_keys]
 
+            # channel_keys does not contain a single redis key more than once
             async with self.connection(connection_index) as connection:
                 await connection.eval(group_send_lua, keys=channel_redis_keys, args=args)
```
gekco: previously channel_redis_keys would contain duplicate keys as multiple channels may map to same redis key. hence the more number of pushes.

andrewgodwin:
OK, I mostly understand it now, but a couple of things I'd like to see changed:

- The `_map_channel_to_connection` function is now far too complex - that whole thing should be refactored so it doesn't have to pass around things as dicts of lists and so forth. Maybe flatten it more into the only place it's called.

- I'm not particularly happy with the comma-separated approach to the storing of channel names. Maybe just use a native msgpack list?

gekco:
1. I have restructured my code a bit , also using a more relevant function name, eliminated some repitive code, flattened it, have a look. i am now using _map_channel_keys_to_connection function.
2. I am using a msgpack native list now

andrewgodwin:
Thanks!

gekco:
@andrewgodwin Thank you so much. This was my first PR to get merged . Yayy. :D

PR #111, opened 2018-07-03
Persist

ianjw11:


jberci:
As the author of some of the commits above, I'd just like to add that this is still a work in progress and our team does intend to finish with a pull request, as we've already commented in issue #100 .

As we've seen internally though, the code above by itself isn't enough because it still leaves open some cancellation problems and related race conditions. We've also uncovered some problems in asyncio (which have happily been fixed in recent upstream) which made debugging difficult for a time.

You can track progress on our temporary fork. Part of the remaining fix is already implemented, the rest of it coming up shortly along with a merge with latest django/channels_redis master.

Side note: I don't believe our truncated README and local metadata should be part of any pull request ;)

andrewgodwin:
Alright, I'll let this sit until you're ready to merge, just let me know if you need some help!

PR #114, opened 2018-07-29
Redis pool persistent connection

gekco:
Regarding #83 #100 
Hi,
This PR makes use of the redis connection pool instead of native redis connection objects.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -208,8 +208,8 @@ class RedisChannelLayer(BaseChannelLayer):
                     await self.receive_buffer[channel].put(message)
             else:
                 await self.receive_buffer[real_channel].put(message)
-            # this sleep prevents makes sure that the redis queue is not popped
-            # before the previous message was processed and returned back to the caller
+            # this sleep makes sure that the redis queue is not popped before the
+            # previous message was processed and returned back to the caller
             await  asyncio.sleep(0.0001)
 
     async def receive_single(self, channel):
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -208,6 +208,9 @@ class RedisChannelLayer(BaseChannelLayer):
                     await self.receive_buffer[channel].put(message)
             else:
                 await self.receive_buffer[real_channel].put(message)
+            # this sleep prevents makes sure that the redis queue is not popped
+            # before the previous message was processed and returned back to the caller
+            await  asyncio.sleep(0.0001)
 
     async def receive_single(self, channel):
         """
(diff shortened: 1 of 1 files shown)


channels_redis/core.py
```diff
@@ -208,6 +208,9 @@ def _setup_encryption(self, symmetric_encryption_keys):
                     await self.receive_buffer[channel].put(message)
             else:
                 await self.receive_buffer[real_channel].put(message)
+            # this sleep makes sure that the redis queue is not popped before the
+            # previous message was processed and returned back to the caller
+            await  asyncio.sleep(0.0001)
```
gekco: i am not happy with this, but i couldn't find any other solution to this. as when i used connection pools , the last test case which was creating an issue:

Test Case  No 16 :
get a new channel - > add channel to group -> send four messages -> one by one call receive

PROBLEM : Alternate messages were getting skipped and following flow of events is being followed
  
receive -> start receive_loop -> wait for pop result -> pop result_1 -> pop result_2 -> process and return result 1 from receive -> CANCEL receive_loop -> call receive again -> pop result_3 -> process and return result_3

result_2 was being popped from the queue and the task was then getting cancelled. let me know if you come up with a better solution.

gekco: This sleep schedules the receive coroutine before again awaiting the receive_single coroutine which in turns pops the redis queue. now the receive_loop was cancelled before even the message was popped from redis.

andrewgodwin:
This results in a dangling coroutine (see the `Task was destroyed but it is pending` output at https://travis-ci.org/django/channels_redis/jobs/409574263) so I can't take it until that's fixed. That's why we had to stop using the pool objects in the first place

gekco:
Ahh I guess i somehow skipped the warning:
Anyhow The reason for that was the receive_loop routine was not getting cancelled immediately on cancel. I have resolved that by awaiting the coroutine after cancelling it. @andrewgodwin please have a look

channels_redis/core.py
```diff
@@ -191,6 +192,12 @@ def _setup_encryption(self, symmetric_encryption_keys):
                 # If we were the last out, stop the receive loop
                 if self.receive_count == 0:
                     self.receive_loop_task.cancel()
+                    # as tasks dont get cancelled immediately we will await the task
+                    # in order to make sure it gets cancelled properly
+                    try:
+                        await self.receive_loop_task
+                    except CancelledError:
+                        pass
```
gekco: as with a task , it does not get cancelled immediately after calling cancel, CancelledError Need sto be raised for that task to cancel. 
[](https://stackoverflow.com/questions/40897428/please-explain-task-was-destroyed-but-it-is-pending)
Due to this when the loop stops after the tests, the task was pending. 
Awaiting it so as to raise the CancelledError and get the task cancelled before the receive loop completes.

channels_redis/core.py
```diff
@@ -208,6 +215,9 @@ def _setup_encryption(self, symmetric_encryption_keys):
                     await self.receive_buffer[channel].put(message)
             else:
                 await self.receive_buffer[real_channel].put(message)
+            # this sleep makes sure that the redis queue is not popped before the
+            # previous message was processed and returned back to the caller
+            await asyncio.sleep(0.0001)
```
andrewgodwin: Why does the sleep do this? Sleeping for 0.1 milliseconds sounds like a hack to ensure something else happens when in reality it should be a lock.

channels_redis/core.py
```diff
@@ -487,18 +497,34 @@ def connection(self, index):
...
+
+        def __init__(self, kwargs, index):
             self.kwargs = kwargs
+            self.index = index
 
         async def __aenter__(self):
-            self.conn = await aioredis.create_redis(**self.kwargs)
-            return self.conn
+            # get connection key for the current loop and the given index
+            connection_key = asyncio.get_event_loop().__hash__() + self.index
```
andrewgodwin: Why hash rather than ID? I would say ID makes more sense since it's an object instance, unless event loops implemented the hash function when I wasn't looking!

channels_redis/core.py
gekco: well the sleep halts this coroutine after pushing the message in the channel receive buffer. so that the control moves to the receive coroutine before running the while loop again and it always does guarantee this, but anyways i am trying to come up with a better way to schedule the receive coroutine after a single pop from redis queue

channels_redis/core.py
gekco: Yup.agreed. will change

PR #117, opened 2018-08-08
Connection pooling and cancellation fixes

jberci:
This pull request contains everything we at @genialis had to do to make connection pooling work. The commits aren't squashed so that the commit logs are preserved; they describe the problems we stumbled upon.

The changes implement event loop-local connection pooling and introduce some cleanup hooks needed to avoid coroutine leaks and the like.

Message receiving also needed some work because of cancellation problems. We've seen locally that some environments cancel requests rather aggressively, which led to massive message loss with the old code. The reworked handling utilises Redis' concept of reliable queues.

Some of the original problems were due to bugs in the asyncio locking implementation, which were only fixed recently. At the time of this writing, the fixes are in Python 3.6.5 but haven't been backported further, so this code requires at least Python 3.6.5.

All tests pass and we also haven't seen any problems internally that could be attributed to this in a while. Please review.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -15,6 +15,81 @@ from channels.exceptions import ChannelFull
 from channels.layers import BaseChannelLayer
 
 
+class ConnectionPool:
+    """
+    Connection pool manager for the channel layer.
+
+    It manages a set of connections for the given host specification and
+    taking into account asyncio event loops.
+    """
+
+    def __init__(self, host):
+        self.host = host
+        self.conn_map = {}
+        self.in_use = {}
+
+    def _ensure_loop(self, loop):
+        """
+        Get connection list for the specified loop.
+        """
+        if loop is None:
+            loop = asyncio.get_event_loop()
+
+        if loop not in self.conn_map:
+            self.conn_map[loop] = []
+
+        return self.conn_map[loop], loop
+
+    async def pop(self, loop=None):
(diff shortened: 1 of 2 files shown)


jberci:
The switch was needed for the reliable queue implementation: Redis only has a `BRPOPLPUSH` combined command, so everything had to be changed to pop from the right and push to the left. See 
https://redis.io/commands/rpoplpush#pattern-reliable-queue .

andrewgodwin:
Aha! That was the key thing I was missing, thanks. That does mean that this will become a backwards-incompatible deploy technically, but honestly the worst thing it will do is reverse message priority on deploy so I'm alright with that, so I'll merge it in.

Thanks for your work on this!

pr4n:
Does it mean that we **finally** have a persistent connection? and that https://github.com/django/channels_redis/pull/114 and https://github.com/django/channels_redis/pull/104 are not needed?

andrewgodwin:
@jberci Have you seen the problems with this raised in https://github.com/django/channels_redis/issues/125? I'm going to have to revert this PR back out if we can't get those solved, unfortunately.

harmv:
NB: connection pooling (https://github.com/django/channels_redis/pull/117) does not work when using async_to_sync() around group_send()

See: https://github.com/django/channels_redis/issues/223#issuecomment-1084688116

PR #128, opened 2018-09-08
PendingDeprecationWarning: encoding is deprecated, Use raw=False instead

kradem:


andrewgodwin:
Do you know what version this changed in and when `raw` was introduced? We might need to update the library version requirement too.

kradem:
Well [here](https://pypi.org/project/msgpack/#history) it was introduced for the first time in the [version 0.5.2](https://pypi.org/project/msgpack/0.5.2/).

andrewgodwin:
Then you will need to update the dependency in `setup.py` to require at least 0.5.2 before I can merge this!

andrewgodwin:
Cool, thanks!

PR #130, opened 2018-09-18
Fix pool management on event loop shutdown

### LINKED ISSUE #125: Redis connections keep increasing when used with async_to_sync
I noticed that our task runner (huey) was keeping over 1000 connections to Redis open. After a day of digging around, I noticed that it can also happen in a plain Django Channels environment.

Here's a simple repro using django management commands:

https://github.com/tiltec/channels-redis-sync-repro/

It shows that sending with `async_to_sync` opens a new redis connection per message and never closes it again.

This is a regression from channels_redis 2.2.1.

### Related issues:
- https://github.com/django/channels_redis/issues/124
- https://github.com/django/channels_redis/issues/122

jberci:
There are very few ways to get notified about when an asyncio event loop
shuts down. The asgiref package makes sure that asynchronous generators
are explicitly cleaned up after a call to async_to_sync, so this patch
uses a dummy async generator to "register" for loop shutdown.
This should fix #125.

Many thanks to @tiltec for providing the repro repo and @ruralscenery for pointing me in the right direction.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -7,12 +7,30 @@ import itertools
 import random
 import string
 import time
+import types
 
 import aioredis
 import msgpack
 
 from channels.exceptions import ChannelFull
 from channels.layers import BaseChannelLayer
+
+
+def _wrap_close(loop, pool):
+    """
+    Decorate an event loop's close method with our own.
+    """
+    original_impl = loop.close
+
+    def _wrapper(self, *args, **kwargs):
+        # If the event loop was closed, there's nothing we can do anymore.
+        if not self.is_closed():
+            self.run_until_complete(pool.close_loop(self))
+        # Restore the original close() implementation after we're done.
+        self.close = original_impl
+        return self.close(*args, **kwargs)
+
+    loop.close = types.MethodType(_wrapper, loop)
 
 
 class ConnectionPool:
@@ -36,6 +54,9 @@ class ConnectionPool:
             loop = asyncio.get_event_loop()
 
         if loop not in self.conn_map:
+            # Swap the loop's close method with our own so we get
+            # a chance to do some cleanup.
+            _wrap_close(loop, self)
             self.conn_map[loop] = []
 
         return self.conn_map[loop], loop
@@ -57,8 +78,9 @@ class ConnectionPool:
         """
         loop = self.in_use[conn]
         del self.in_use[conn]
-        conns, _ = self._ensure_loop(loop)
-        conns.append(conn)
+        if loop is not None:
+            conns, _ = self._ensure_loop(loop)
+            conns.append(conn)
 
     def conn_error(self, conn):
         """
@@ -73,6 +95,20 @@ class ConnectionPool:
         """
         self.conn_map = {}
         self.in_use = {}
+
+    async def close_loop(self, loop):
+        """
+        Close all connections owned by the pool on the given loop.
+        """
+        if loop in self.conn_map:
+            for conn in self.conn_map[loop]:
+                conn.close()
+                await conn.wait_closed()
+            del self.conn_map[loop]
+
+        for k, v in self.in_use.items():
+            if v is loop:
+                self.in_use[k] = None
 
     async def close(self):
         """
@@ -732,7 +768,7 @@ class RedisChannelLayer(BaseChannelLayer):
         """
         if isinstance(value, str):
             value = value.encode("utf8")
-        bigval = binascii.crc32(value) & 0xfff
+        bigval = binascii.crc32(value) & 0xFFF
         ring_divisor = 4096 / float(self.ring_size)
         return int(bigval / ring_divisor)
 


andrewgodwin:
Looks like this won't work on Python 3.5 as written as you've got a `yield` inside an `async def`.

jberci:
As far as I can see, Python 3.5 simply has no facilities that would enable any kind of clean event loop shutdown. We could hack our way through by decorating the `close` method of any event loop that passes through the pool and that decorator would then do the cleanup and call the original implementation. asgiref does call `close` on all the loops it creates internally, so this should actually work, but it feels like a dirty approach to me, that's why I went with the async generator idea.

If you'd prefer, I can implement that as well and make another pull request. In that case, people would be able to use channels_redis on Python 3.5... but they'd still be left with all the asyncio synchronization bugs which affect both the pool and pre-pool variants of channels_redis and probably won't get fixed. As #117 already mentioned, 3.6.5 is the lowest Python version with all the fixes.

andrewgodwin:
I'm alright with making a version that works on 3.5 but isn't perfect, as long as it's not a regression, and then we can make it very clear that 3.6 is strongly preferred.

pythonpro:
Guys, pls check this solution by doing several thousands of operations in a row.
I've checked it in my project. It works at the beginning, but after several thousands of
async_to_sync(channel_layer.group_send)(...)
calls it throws the exception:
-- trim --
`File "/usr/lib/python3.7/asyncio/base_events.py", line 943, in create_connection
    raise exceptions[0]
  File "/usr/lib/python3.7/asyncio/base_events.py", line 930, in create_connection
    await self.sock_connect(sock, address)
  File "/usr/lib/python3.7/asyncio/selector_events.py", line 475, in sock_connect
    return await fut
  File "/usr/lib/python3.7/asyncio/selector_events.py", line 480, in _sock_connect
    sock.connect(address)
OSError: [Errno 99] Cannot assign requested address`

ztaylor4:
I have had a similar error in Python 3.6.5:

Code (simplified):

    async_to_sync(channel_layer.group_send)('controller', {
        "type": "send_controller_data",
    })

Traceback (trimmed at the top):

    File "/home/utilitygg/.virtualenvs/controller/lib/python3.6/site-packages/asgiref/sync.py", line 64, in __call__
      return call_result.result()
    File "/usr/lib/python3.6/concurrent/futures/_base.py", line 425, in result
      return self.__get_result()
    File "/usr/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result
      raise self._exception
    File "/home/utilitygg/.virtualenvs/controller/lib/python3.6/site-packages/asgiref/sync.py", line 78, in main_wrap
      result = await self.awaitable(*args, **kwargs)
    File "/home/utilitygg/.virtualenvs/controller/lib/python3.6/site-packages/channels_redis/core.py", line 559, in group_send
      async with self.connection(self.consistent_hash(group)) as connection:
    File "/home/utilitygg/.virtualenvs/controller/lib/python3.6/site-packages/channels_redis/core.py", line 742, in __aenter__
      self.conn = await self.pool.pop()
    File "/home/utilitygg/.virtualenvs/controller/lib/python3.6/site-packages/channels_redis/core.py", line 49, in pop
      conns.append(await aioredis.create_redis(**self.host, loop=loop))
    File "/home/utilitygg/.virtualenvs/controller/lib/python3.6/site-packages/aioredis/commands/__init__.py", line 177, in create_redis
      loop=loop)
    File "/home/utilitygg/.virtualenvs/controller/lib/python3.6/site-packages/aioredis/connection.py", line 107, in create_connection
      timeout, loop=loop)
    File "/usr/lib/python3.6/asyncio/tasks.py", line 339, in wait_for
      return (yield from fut)
    File "/home/utilitygg/.virtualenvs/controller/lib/python3.6/site-packages/aioredis/stream.py", line 19, in open_connection
      lambda: protocol, host, port, **kwds)
    File "/usr/lib/python3.6/asyncio/base_events.py", line 778, in create_connection
      raise exceptions[0]
    File "/usr/lib/python3.6/asyncio/base_events.py", line 765, in create_connection
      yield from self.sock_connect(sock, address)
    File "/usr/lib/python3.6/asyncio/selector_events.py", line 450, in sock_connect
      return (yield from fut)
    File "/usr/lib/python3.6/asyncio/selector_events.py", line 455, in _sock_connect
      sock.connect(address)
    
    OSError: [Errno 99] Cannot assign requested address

After increasing my file descriptor ulimit, this error shows up after a couple of hours.  To induce the error more quickly, I loop the code above to get called very often and monitor the increasing number of open files with the code below:

    watch -n10 "sudo lsof -w | cut -d \" \" -f 1 | uniq -c | sort -nr | head -n15"

Relevant pip:

    aioredis==1.1.0
    asgiref==2.3.2
    channels==2.1.3
    channels-redis==2.3.0
    Django==2.0.4
    redis==2.10.6

jberci:
From what you've pasted, you might be running into a system-defined limit. Every TCP connection (such as the ones used by channels_redis) will always have a system-defined timeout period from when you shut it down cleanly to when the system actually frees resources associated with it. Usually, this will be around 1 minute.

Before connection pooling, this was a problem as soon as you sent many messages through the layer, no matter what your design was. With connection pooling, this shouldn't happen at all anymore in async environments.

For sync environments, all we can do is make sure connections aren't leaked, which is what this pull request does. Sending messages in tight loops will still cause many connections to be created because each connection is tied to a specific event loop and can't be reused in others (this is due to asyncio design and can't be fixed  in channels_redis).

To avoid running into system limits, you can either avoid sending many messages at once with `async_to_sync`, or make sure to do that in an environment where an  event loop is already available, so that asgiref won't create a new one for each call.

jberci:
@andrewgodwin I've updated the commit with the alternative fix. It's not as conceptually clean as the previous one, but it is version-independent and there is less code. It works with the test suite as well as with the reproducer from the original issue.

channels_redis/core.py
```diff
@@ -36,6 +53,7 @@ def _ensure_loop(self, loop):
             loop = asyncio.get_event_loop()
 
         if loop not in self.conn_map:
+            _wrap_close(loop, self)
```
andrewgodwin: Let's highlight with a comment what's happening here

jberci: Done. There's also an unrelated change  in `consistent_hash` to make the linter happy.

andrewgodwin:
Thanks so much - and yes, Black appears to have got an update recently, so the linting jumped slightly there.

ploggingdev:
The 2.3.0 version available in pypi keeps crashing. Looks like this PR fixes the issue, so could we get a release?

andrewgodwin:
I’m both ill and travelling for the next two weeks so a release might take a bit as I need to manually verify the fixes. If you want to run the master branch and verify it works that would help my confidence level.

andrewgodwin:
2.3.1 is now released with this in it.

chrismeyersfsu:
Instead of using asgiref's `async_to_sync` you can use the simple implementation below.

```
def run_sync(func):
    event_loop = None
    try:
        event_loop = asyncio.get_event_loop()
    except RuntimeError:
        event_loop = asyncio.new_event_loop()
        asyncio.set_event_loop(event_loop)
    return event_loop.run_until_complete(func)
```

Issue #136, opened 2018-11-07
Locking exception gets thrown on occasion


maliesa96:
I'm using channels for a personal project and every now and then an exception gets thrown regarding locking. It doesn't seem to be caused by any action in particular, it just seems to happen randomly. I can't seem to find a solution. 

I'm using Python 3.7
Traceback:

Exception inside application: 
  File "/Users/Mishari/apollo-env/lib/python3.7/site-packages/channels/sessions.py", line 179, in __call__
    return await self.inner(receive, self.send)
  File "/Users/Mishari/apollo-env/lib/python3.7/site-packages/channels/consumer.py", line 59, in __call__
    [receive, self.channel_receive], self.dispatch
  File "/Users/Mishari/apollo-env/lib/python3.7/site-packages/channels/utils.py", line 59, in await_many_dispatch
    await task
  File "/Users/Mishari/apollo-env/lib/python3.7/site-packages/channels_redis/core.py", line 449, in receive
    assert not self.receive_lock.locked()

freeze:
aioredis==1.2.0
asgiref==2.3.2
async-timeout==2.0.1
attrs==18.2.0
autobahn==18.9.2
Automat==0.7.0
channels==2.1.4
channels-redis==2.3.1
constantly==15.1.0
daphne==2.2.2
Django==2.1.2
hiredis==0.2.0
hyperlink==18.0.0
idna==2.7
incremental==17.5.0
msgpack==0.5.6
PyHamcrest==1.9.0
pytz==2018.5
six==1.11.0
Twisted==18.9.0
txaio==18.8.1
zope.interface==4.5.0

Thanks in advance

andrewgodwin:
I've heard reports of this occasionally from elsewhere, but I need more information to solve this - specifically, I need a way to reliably reproduce it. Without that, there's not much we can do!

maliesa96:
I'm thinking to use pdb to see if I can get you some more info. I'll update if i find anything.

maliesa96:
So it seems to happen after a websocket disconnects. I got pdb to trigger a breakpoint. I can see that
`self.receive_lock.locked()` evaluates to true when it shouldn't be (I assume that was pretty obvious).

Could this possibly be due to the fact that the websocket was closed abruptly without triggering the disconnect function in the consumer?

andrewgodwin:
Maybe - we might be missing a `finally` somewhere that isn't getting triggered on the way out of an abrupt disconnect. That locking code is... complicated.

[...]

Issue #134, opened 2018-02-28
Multiple Daphne instances with Redis backend get stuck


ralphje:
I am currently debugging the following situation, for which I have not been able to create a reproducible situation in a testing environment, mostly due to time contraints. I am posting what I know now, perhaps this rings a bell somewhere that makes me stop looking further, but I suspect a bug of some kind:

- Chromium browser client, Ubuntu 16.04 server
- channels 2.0.2, daphne 2.0.4, Django 2.0.2, channels_redis 2.1.0

I have a Nginx proxy server that proxies between two Daphne instances running the same codebase. I am running a Redis channel backend. I am using Celery for several backend tasks, and want to report their status to the client that is waiting on them. For this, they publish a PROGRESS state and I have hooked some Channels code inside the storage function, so that a message is sent to a group waiting on that task, e.g. 

```python
async_to_sync(channel_layer.group_send)("celery-status-<task-uuid>", {"type": "celery.task_status", "text": task_status})
```

I have a consumer that looks a bit like this:

```python
class MyConsumer(AsyncWebsocketConsumer):
    async def receive(self, text_data=None, bytes_data=None):
        await self.channel_layer.group_add("celery-status-<task-uuid>", self.channel_name)

    async def celery_task_status(self, event):
        await self.send(text_data=event["text"])
```

Now, when someone requests to receive updates for the given Celery task ID, and I send status updates to the client. (I also immediately send the current status to the consumer, which works fine, so is left out of the code example).

The bug I encounter, is that after a while, progress updates are not sent to the clients anymore. This only appears to happen when there are two or more Daphne servers running. The Redis queues fill up. When I shutdown one of those Daphne servers, and clear the Redis cache, it continues to work fine. 

Perhaps this has to do with my Nginx proxy setup, which has simply two upstream servers specified, although I'm not sure that's the culprit. I think two Daphne instances is the root of the problem.

Is this a known issue?
Is there something I can add to this bug to make it more reproducible?

khpeterson:
I am running into what seems like the same issue - multiple instances of daphne running behind apache on AWS, communicating through ElastiCache/redis.  The clients consist of simple websocket event producer/consumer pairs.

[...]

PR #140, opened 2018-11-25
release receive_lock on CancelledError when task.cancel() fails

khpeterson:
This fix resolves the stuck channel layer I bumped into in my AWS production environment.  I've been running there with this fix for several days now (with websocket_timeout=300 to exacerbate the issue) and haven't seen a single ChannelFull exception or self.receive_lock.locked() assertion. 

I posted about my findings in https://github.com/django/channels_redis/issues/134 and I believe the fix also addresses https://github.com/django/channels_redis/issues/136.

The test case I created in test_core.py is a little dubious because it's timing dependent but I'm not sure how else to tickle the race condition.  Suggestions welcome.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -384,8 +384,7 @@ class RedisChannelLayer(BaseChannelLayer):
                         # Ensure all tasks are cancelled if we are cancelled.
                         # Also see: https://bugs.python.org/issue23859
                         for task in tasks:
-                            cancelOK = task.cancel()
-                            if not cancelOK:
+                            if not task.cancel():
                                 assert task.done()
                                 if task.result() is True:
                                     self.receive_lock.release()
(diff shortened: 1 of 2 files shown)


channels_redis/core.py
```diff
@@ -384,7 +384,11 @@ def _backup_channel_name(self, channel):
                         # Ensure all tasks are cancelled if we are cancelled.
                         # Also see: https://bugs.python.org/issue23859
                         for task in tasks:
-                            task.cancel()
+                            cancelOK = task.cancel()
+                            if not cancelOK:
```
andrewgodwin: You can combine `task.cancel` into the `if not` statement here and avoid the variable

khpeterson: Thanks Andrew, cleanups committed.

andrewgodwin:
Thanks a bunch! Hopefully this will fix #136 too - I will post about it there.

PR #142, opened 2018-11-28
Add 'python_requires' check

mikemahony:
This will prevent users of Python <3.5.2 from installing the package.


andrewgodwin:
Thanks!

PR #153, opened 2019-04-11
Bump asgiref to 3.0

michael-k:
asgiref < 3 does not work with the latest daphne release and won't work with the next channels release.  See django/channels#1277

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,8 +33,8 @@ setup(
     install_requires=[
         "aioredis~=1.0",
         "msgpack~=0.6.0",
-        "asgiref~=2.1",
-        "channels~=2.0",
+        "asgiref~=3.0",
+        "channels~=2.2",
     ],
     extras_require={"cryptography": crypto_requires, "tests": test_requires},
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -33,7 +33,7 @@
     install_requires=[
         "aioredis~=1.0",
         "msgpack~=0.6.0",
-        "asgiref~=2.1",
+        "asgiref~=3.0",
         "channels~=2.0",
```
michael-k: channels should be bumped as well, because channels >=2.0,<2.1.8 does not support asgiref ~=3.0.
I cannot bump this yet as I don't know the version of channel's next release.

carltongibson:
But, of course, I'll need to release it before we can run the CI here... 🙂

daxaxelrod:
Thanks for bumping and releasing guys :D Just a quick note in case someone runs into this in the future. You need python >=3.6. My pipenv was set up with 3.5 which left made me upgrade everything

Issue #151, opened 2019-04-04
Wrong order of messages in process specific channel if receive call was cancelled.


sergey-komissarov:
Hello, we are using group to send notifications from channels worker process to the websocket consumer in the server process. There is exactly one worker in the group and exactly one consumer on the server side. Sometimes messages order is changed on the server side. 

Detailed investigation shows the following events on the server side:
1. For the channel `specific.innxCCgU!NcMqIZztkRQJ` `receive_lock` acquired and `receive_single` started.
2. Before this call [core.py#L483](https://github.com/django/channels_redis/blob/master/channels_redis/core.py#L483) the redis channel messages list had 3 messages and backup list was empty:
```
asgi:specific.innxCCgU! = [msg3, msg2, msg1]
asgi:specific.innxCCgU!$inflight = []
```
3. Receive was cancelled and `receive_lock` was released:
```
Traceback (most recent call last):
  File "/pSeven/venv/lib/python3.7/site-packages/channels_redis/core.py", line 429, in receive
    real_channel
  File "/pSeven/venv/lib/python3.7/site-packages/channels_redis/core.py", line 484, in receive_single
    index, channel_key, timeout=self.brpop_timeout
  File "/pSeven/venv/lib/python3.7/site-packages/channels_redis/core.py", line 329, in _brpop_with_clean
    return await connection.brpoplpush(channel, backup_queue, timeout=timeout)
concurrent.futures._base.CancelledError
```
4. The next call to receive for the channel `specific.innxCCgU!EKHzfqiTYmHm` acquired `receive_lock` and found redis lists in the following state:
```
asgi:specific.innxCCgU! = [msg3, msg2]
asgi:specific.innxCCgU!$inflight = [msg1]
```
It seems that the redis server already moved message to the backup list during `brpoplpush` operation but the message body was not fully received when the task was cancelled.
5. After [core.py#L327](https://github.com/django/channels_redis/blob/master/channels_redis/core.py#L327) `msg1` is placed to the beginning of the list `asgi:specific.innxCCgU!` and `msg2` is returned by receive.
6. Next call to the receive will return `msg3` and only the next will get `msg1` from the redis.

[...]

PR #165, opened 2019-07-26
Rewrites RedisChannelLayer.receive specific channel section

augonis:
I had a mystery bug where the channel_layer would stop receiving on specific channels after a bunch of receives got canceled.
Unable to reproduce the bug in clean conditions and unable to figure out what exactly was wrong, I just rewrote the suspect (`RedisChannelLayer.receive`).
My bug is gone and the tests still pass, so the pull request.

As a bonus, now simultaneously receiving on '!' channels with different prefixes should work (e.g. channel names produced by `redisChannelLayer.new_channel(prefix='something')`), witch was not the case before.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -166,6 +166,66 @@ class UnsupportedRedis(Exception):
     pass
 
 
+class ReceiveBuffer:
+    def __init__(self, channel_layer):
+        self.loop = None
+        self.channel_layer = channel_layer
+        self.getters = collections.defaultdict(collections.deque)
+        self.buffers = collections.defaultdict(lambda: collections.deque(maxlen=20))
+        self.receiver = None
+    
+    def __bool__(self):
+        return bool(self.getters)
+    
+    def get(self, channel):
+        getter = self.loop.create_future()
+        
+        if channel in self.buffers:
+            getter.set_result(self.buffers[channel].popleft())
+            if not self.buffers[channel]:
+                del self.buffers[channel]
+        else:
+            getter.channel = channel
+            getter.add_done_callback(self._getter_done_prematurely)
+            self.getters[channel].append(getter)
+            
+            # ensure receiver is running
+            if not self.receiver:
+                self.receiver = asyncio.ensure_future(self.receiver_factory(self.channel_layer.non_local_name(channel)))
+        
+        return getter
+    
+    def _getter_done_prematurely(self, getter):
+        channel = getter.channel
+        self.getters[channel].remove(getter)
+        if not self.getters[channel]:
+            del self.getters[channel]
+        if not self and self.receiver:
+            self.receiver.cancel()
+    
+    def put(self, channel, message):
+        if channel in self.getters:
+            getter = self.getters[channel].popleft()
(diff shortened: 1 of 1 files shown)


carltongibson:
Hi @augonis. Interesting... 

So, tests...  

What’s our new buffer meant to offer that `Queue` doesn’t? Can we add some tests showing that in action?

augonis:
The ReceiveBuffer is much more than a `Queue`, it manages waiters and buffers messages for all specific channels under the same 'real channel', as well as runs the receive loop for the 'real channel'.
It does have some Queue like bits in its implementation. 
Only useful exposed are `.loop` and `.get(channel`), rest is implementation.

My goal was to simplify the control logic with the locks for the receiving as I suspect something there was behind the mystery bug.
Instead of locks I use a worker `Task`/loop that `.single_receive(real_channel)` that puts message in `ReceiveBuffer` that then either stores the massage for later or fulfills a getter `Future`.

I'll work on some tests.

augonis:
I added tests for the receive task handling, as basic receive is covered by existing tests.

P.S. I've put a limit on how many messages per specific channel `receive Buffer` will hold (hardcoded 20) unlike the original `.receive` that would buffer unlimited messages. Any thoughts?

carltongibson:
> P.S. I've put a limit on how many messages per specific channel receive Buffer will hold (hardcoded 20) unlike the original .receive that would buffer unlimited messages. Any thoughts?

I wondering if we can tie this into `capacity`. (Then it's directly configurable...)

Overall this looks nice. It's my summer project to put together a set of decent releases for Daphne, Channels and here, so let's get this in. Thanks for the input! 🏅

augonis:
Doable, but semantically and implementation wise it would muddy the meaning of `capacity` (and friends `channel_capacity`, `get_capacity(channel)`).
* Currently  `capacity` is configured/enforced on the sending side, having it affect receiving would be something new.
* `capacity` also applies per real channel rather than per specific channel, so how it relates would have to be resolved

carltongibson:
Yes... but it should be exposed, and does it merit another config option? (Not sure...)

augonis:
Should I remove the limit to unblock this PR and open a separate issue?

augonis:
Just read #134, sounds similar my problem, this PR might address it.
#151 also may be related if that issue is only with specific channels.

qeternity:
What's the status with this?

carltongibson:
It just needs some time to review properly. I have some issues around Channels' AsyncHttpConsumer at the top of my list, and then channels_redis after that.

davidfstr:
`self.receive_buffers` is added to but never removed from. Memory leak. Probably want to dereference receive buffers that become empty.

There's also the previously-existing memory leak related to channels that are written to but never written from, in https://github.com/django/channels/issues/1361 , which I don't advocate fixing in this commit because it's already big enough.

augonis:
I'd expect receive_buffers to be long lived and few in number.
In fact for all existing applications there would be just one, as receiving on multiple `real_channels` previously was not supported, though implied by the ability to specify prefix in `redisChannelLayer.new_channel(prefix='something')`.

PR #166, opened 2019-08-08
Asgi 1.5

xrg:
DO NOT merge.

Please, can we have this HEAD as a **new** "1.5" branch?
It will allow legacy projects, using channels 1.0 to use redis >=3.2.x , bridge to python3 compatibility.

### DIFF
--- a/asgi_redis/__init__.py
+++ b/asgi_redis/__init__.py
@@ -2,4 +2,4 @@ from .core import RedisChannelLayer
 from .local import RedisLocalChannelLayer
 from .sentinel import RedisSentinelChannelLayer
 
-__version__ = '1.4.3'
+__version__ = '1.5.0'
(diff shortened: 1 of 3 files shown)


carltongibson:
Hi @xrg. Hmmm. 

I have zero capacity to maintain an old branch here. I mean I can push this to a branch sure but, how does that help you vs you just maintaining the same in your own fork?

xrg:
@carltongibson I can maintain that branch, no problem. My only concern with that PR was to /reserve/ that version number, get your consent about it, so that it won't get used for any other fork/branch ever. Is that ok with you?

carltongibson:
Hi @xrg, I certainly won't add a `1.5` branch here. I'm happy for you to push that to your fork. (I have no control over, or other forks, so I can't stop it being used on other forks ever.) 

I am reluctant to push such a branch here, as it's not something I can allocate resources to reviewing.

PR #171, opened 2019-10-21
Support Python 3.8

### LINKED ISSUE #170: Doesn't work with Python 3.8
Suddenly stops working after upgrade from Python 3.7.3 to Python 3.8.0.
Consumers don't get group_send messages. No exceptions of any type.  
Downgrade to Python 3.7.3 solves the problem.

- Docker runtime
- channels==2.3.0
- channels-redis==2.4.0
- django==2.2.6
- Redis version - 5.0.5

laevilgenius:
`asyncio.CancelledError` is a subclass of `BaseException` [now](https://bugs.python.org/issue32528).

Potentially fixes #170.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ sudo: false
 language: python
 
 python:
+  - "3.8"
+  - "3.7"
   - "3.6"
 
 services:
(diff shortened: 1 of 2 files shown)


carltongibson:
@laevilgenius Why is this draft? I'd merge it now. Thanks.

laevilgenius:
I've been waiting for `asgiref` version bump. Now that it is released, I'll update `channels` PR, so we can bump it here as well.

PR #177, opened 2020-01-18
Fix double colon in group names.

chouhanyang:
Remove the extra colon since prefix already contains a colon at the end.

e.g. "asgi::group:ABB47B09AA1C9425"

Tested with my own build. Made sure the extra colon was removed.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -615,9 +615,11 @@ class RedisChannelLayer(BaseChannelLayer):
                 x.decode("utf8") for x in await connection.zrange(key, 0, -1)
             ]
 
-        connection_to_channel_keys, channel_keys_to_message, channel_keys_to_capacity = self._map_channel_keys_to_connection(
-            channel_names, message
-        )
+        (
+            connection_to_channel_keys,
+            channel_keys_to_message,
+            channel_keys_to_capacity,
+        ) = self._map_channel_keys_to_connection(channel_names, message)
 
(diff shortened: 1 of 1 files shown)


chouhanyang:
Weird. Not clear to me why the checks failed. Any pointers?

chouhanyang:
Thanks for the poniter @carltongibson . 

Originally, I just wanted to make a minimal change. Now as you suggested, I put a separate change on top of it to normalize all prefixes. Now that:

1. All prefix default values contain trailing separators (that is to conform new_channel(self, prefix="specific.") in BaseChannelLayer), which ironically has the same problem.

```
    async def new_channel(self, prefix="specific."):
        """
        Returns a new channel name that can be used by something in our
        process as a specific channel.
        """
        return "%s.inmemory!%s" % (
            prefix,
            "".join(random.choice(string.ascii_letters) for i in range(12)),
        )
```

2. All internal prefixes (including client_prefix) has no trailing separator.
3. Add _prefix() to handle separator joining.

Please let me know if this is what in your mind.

carltongibson:
Hi @chouhanyang. 

I think I preferred the more minimal version. 

Perhaps just remove the `:` from the default — would that not be sufficient? 

(My only concern is breaking existing names.)

chouhanyang:
OK, reverted my change to just fix the colon.

carltongibson:
Hi @chouhanyang. Thanks for the input here. I've gone for just removing the `:` from the default in #201.

PR #178, opened 2020-01-24
ConnectionPool.pop must ensure active connections are returned.

### LINKED ISSUE #164: ConnectionClosedError when layer is idle for longer than server timeout
The layer seems to fail when the Redis server is configured to have a timeout and the connection is kept idle for longer than that timeout. This is an issue when using `channels_redis` with a provider that uses a timeout in their instances (for example, Redis To Go sets a timeout of 150 seconds).

To reproduce, start a Redis server with a low timeout value set in `redis.conf`, say 20 seconds. Then run this code:

```python
import asyncio

from channels_redis.core import RedisChannelLayer


async def wait_and_send():
    layer = RedisChannelLayer(hosts=['redis://localhost:6379'])

    # should create new connection
    await layer.group_send('group', {'message': 'hi'})

    # wait for >timeout seconds
    await asyncio.sleep(25)

    # raises exception
    await layer.group_send('group', {'message': 'hi'})


loop = asyncio.get_event_loop()
loop.run_until_complete(wait_and_send())
``` 

When I run it I get:

```python
  File "break_channels_redis.py", line 14, in <module>
    loop.run_until_complete(wait_and_send())
  File "/usr/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "break_channels_redis.py", line 10, in wait_and_send
    await layer.group_send('group', {'message': 'hi'})  # raises exception
  File "/usr/local/lib/python3.7/site-packages/channels_redis/core.py", line 610, in group_send
    key, min=0, max=int(time.time()) - self.group_expiry
  File "/usr/local/lib/python3.7/site-packages/aioredis/commands/sorted_set.py", line 268, in zremrangebyscore
    return self.execute(b'ZREMRANGEBYSCORE', key, min, max)
  File "/usr/local/lib/python3.7/site-packages/aioredis/commands/__init__.py", line 51, in execute
    return self._pool_or_conn.execute(command, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/aioredis/connection.py", line 322, in execute
    raise ConnectionClosedError(msg)
aioredis.errors.ConnectionClosedError: Reader at end of file
```

Following #26 I tried passing `socket_timeout` and `socket_keepalive` as query strings in the Redis URI but they don't seem to have any effect. #135 might also be related.

Rusox89:
Fixes #164

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -618,9 +618,11 @@ class RedisChannelLayer(BaseChannelLayer):
                 x.decode("utf8") for x in await connection.zrange(key, 0, -1)
             ]
 
-        connection_to_channel_keys, channel_keys_to_message, channel_keys_to_capacity = self._map_channel_keys_to_connection(
-            channel_names, message
-        )
+        (
+            connection_to_channel_keys,
+            channel_keys_to_message,
+            channel_keys_to_capacity,
+        ) = self._map_channel_keys_to_connection(channel_names, message)
 
         for connection_index, channel_redis_keys in connection_to_channel_keys.items():
 
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -69,6 +69,9 @@ class ConnectionPool:
         if not conns:
             conns.append(await aioredis.create_redis(**self.host, loop=loop))
         conn = conns.pop()
+        if conn.closed:
+            conn = await self.pop(loop=loop)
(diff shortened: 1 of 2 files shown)


carltongibson:
Hi @Rusox89 Any chance if a regression test for this?

Issue #179, opened 2020-02-09
aioredis 1.3.1 dropped explicit loop requirement in API


edelvalle:
### Your OS and runtime environment, and browser if applicable

 Arch Linux, Python 3.8.1

### What you expected to happen vs. what actually happened?

I expect no warnings, but the dependency aioredis when used in version 1.3.1 is issuing a warning in python 3.8+ that the `loop` parameter is deprecated.

See: https://github.com/aio-libs/aioredis/pull/666

### How you're running Channels (runserver? daphne/runworker? Nginx/Apache in front?):

I'm using `runserver` and `uvicorn`.

### Console logs and full tracebacks of any errors

```
  /home/xigurat/.virtualenvs/iskra/lib/python3.8/site-packages/aioredis/connection.py:105: DeprecationWarning: The loop argument is deprecated
    warnings.warn("The loop argument is deprecated",
```

PR #180, opened 2020-02-09
Trying to slove #179 (aioredis 1.3.1 dropped explicit loop requirement in API)

edelvalle:
Refs #179

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ build/
 *.pyc
 /.tox
 .DS_Store
+.pytest_cache/
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ jobs:
       install: pip install -U .[tests] black pyflakes isort
       script:
         - pyflakes .
-        - black --check .
+        - black --diff --check .
         - isort --check-only --diff --recursive channels_redis tests
 
     - stage: release
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -70,7 +70,7 @@ class ConnectionPool:
         """
         conns, loop = self._ensure_loop(loop)
         if not conns:
-            conns.append(await aioredis.create_redis(**self.host, loop=loop))
+            conns.append(await aioredis.create_redis(**self.host))
         conn = conns.pop()
         if conn.closed:
             conn = await self.pop(loop=loop)
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -1,4 +1,3 @@
-import sys
 import asyncio
 import base64
 import binascii
@@ -8,6 +7,7 @@ import itertools
 import logging
 import random
 import string
+import sys
 import time
 import types
 
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -19,6 +19,8 @@ from channels.layers import BaseChannelLayer
 
 logger = logging.getLogger(__name__)
 
+AIOREDIS_VERSION = tuple(map(int, aioredis.__version__.split('.')))
+
 
 def _wrap_close(loop, pool):
     """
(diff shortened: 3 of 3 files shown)


bjd183:
It looks like you need to fix the linter error by moving `import sys` into alphabetical order (after `import string`).

edelvalle:
Nice! thanks!

bjd183:
`aioredis` implements similar logic [here](https://github.com/aio-libs/aioredis/commit/810651e485ed8d04b7a23f221a0dcc2febd9a85b). I don't know if this is the right approach or not. The loop parameter is deprecated in Python 3.8 where it now automatically determines the appropriate loop (I think). Simply dropping the loop parameter is likely to break Python 3.6-3.7.

bjd183:
Would it be preferred to check the version of aioredis?

edelvalle:
I think you are right, is a combination of the version of python and `ioredis`. I have to check both.

right?

bjd183:
@edelvalle  Perhaps `if sys.version_info >= (3, 8, 0) and aioredis.__version__ >= 1.3.1:` will be acceptable.

edelvalle:
@bjd183 thanks for the advice, I was busy but finally managed to do as instructed by you.
btw, also made `black` tell what's the problem when it gets angry with our code :smile:

bjd183:
@carltongibson how does this look?

carltongibson:
Hi both, thanks for the work here. Looks good. 

Currently in a tight spot. 🦠 disrupting pretty much everything, so I've had ≈0 time to look at anything. I have a commitment to DRF on my list and then I'll have the capacity for Channels &co. 

If you have a bit of time to look over the other issues/PRs (maybe with @tarikki, who's been active) then I'm happy to take group consensus on issues and we can put out a release as first job back. (Equally, no stress if you don't have such capacity.)

bjd183:
What is the best way to reduce to two commits? Squash usually results in a single commit, right? Can this be done with rebase? Or does it require cherry-pick? And are any of these possible now that the changes have already been committed and pushed?

bjd183:
Also, (I've not done this before) is it possible for me to contribute to a pull request that I do not own? How would I do so? It is not obvious to me via this interface.

bjd183:
@edelvalle you could also revert the CI change to black and limit this PR to the fix so it can be easily squashed.

wmorrell:
@bjd183 How I would split one commit into two:

1. Checkout branch containing the commit to split: `git checkout edelvalle/master`
2. Make new branch from the original: `git checkout -b gh-179`
3. Rewind the commit, so `HEAD` is at the parent and the changes in the commit are in working copy: `git reset HEAD~`
4. Add in changes for first commit with: `git add` and/or `git add -p`
5. Do the first commit: `git commit`
6. Add in changes for second commit: `git add`
7. Do the second commit: `git commit`
8. Push
9. Beverage

carltongibson:
Thanks @wmorrell. (Been knocked sideways by the current situation so had little time...)

bjd183:
Alright, I gave that a try. I had to fork `edelvalle/master` since I cannot commit directly to it. I was able to rewind and reset to squash commits to just two. But I wasn't able to push without an additional merge commit since I was then behind the remote tracking branch. I created a PR on `edelvalle/master` which, once merged, I assume propagates to this PR. I'm curious to see how this works since the original set of multiple commits still exists in the history.

edelvalle:
This is worse, now I have even more commits, let me try to undo all this

wmorrell:
Oops, I should have mentioned that it's usually cleaner to re-submit a pull request than to alter (someone else's) existing request. I tried to go for that by adding instruction to make a new branch, but failed to take the suggestion all the way to new PR. I also should have suggested squashing before the `git reset HEAD~`, as it doesn't work very well to reset last commit when the upstream has a small handful of commits instead of one :)

@edelvalle, I will do a fork a bit later to get just the two commits requested. If you would like to give it a shot, I suggest use of a few rounds of `git rebase -i origin/master` along with the `git reset HEAD~` and `git add -p` tools. I'll keep a log of the commands I use, so hopefully it helps someone in the future.

bjd183:
@wmorrell Thanks for the clarification.

carltongibson:
Closing in favour of #193. Thanks all!

PR #183, opened 2020-03-06
Lower logger error on failed group_send from exception to info

tarikki:
Lower logger error from exception to info on failed group send to avoid logger spam.

- [x] Covered by tests.


jberends:
We confirm that the log level should be lowered when the channels is over capacity to info and not raise an exception.

tarikki:
@carltongibson: Let's bring this to the new release to avoid remove logger spam.

carltongibson:
Rebased in #199

PR #184, opened 2020-03-06
Using a sorted set to guarantee message expiration

### LINKED ISSUE #182: Potentially Incorrect Expiration Mechanisms
I'm having tons of problems with channels on my Redis at the moment, after studying carefully the problem I realized that the expiration mechanisms don't work as I expected, first lemme quote the Django Channels specification so we're on the same page expectations-wise:

> Channel layers do not need to persist data long-term; group memberships only need to live as long as a connection
does, and messages only as long as the message expiry time, which is usually a couple of minutes.

### Message expiration
From what I saw the message backlog is saved as asgi:specific.$random12 and the TTL of that key is 60 seconds, and its expiry is being reset for every message (https://github.com/django/channels_redis/blob/master/channels_redis/core.py#L306). This leads to the behavior that messages actually last an infinite amount of time as long as there are new messages being sent, not the 60 seconds per message I expected after reading the documentation and all the READMEs.

My problem: a connection isn't cleanly closed so the group membership is not discarded, leaving that connection with a huge backlog (I've set the capacity to 100000, everything else is the default) that only expires after a day (not 60 seconds). This is consuming a lot of Redis memory in my case and even with a default channel capacity can potentially lead to a mild DoS.

I'm using:
channels-redis 2.4.2
channels 2.4.0
django 3.0.4

My structure:
Redis <- Uvicorn <- Gunicorn <- Nginx <- AWS ELB <- AWS CloudFront

Are my assumptions and expectations correct? If so I will start doing some code for channel_redis. Thanks!!

astutejoe:
Fixes #182 

I went with getrandbits as it generated the 12 random bytes 1 million times in 0.65 seconds against UUID4 that took 3.75 seconds to achieve the same (the difference being UUID4 yields 16 bytes)
I went with 12 bytes to be more memory efficient and still guarantee an astonishing amount of uniqueness.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -5,6 +5,7 @@ import collections
 import hashlib
 import itertools
 import logging
+import random
 import sys
 import time
 import types
@@ -300,12 +301,18 @@ class RedisChannelLayer(BaseChannelLayer):
         else:
             index = next(self._send_index_generator)
         async with self.connection(index) as connection:
+            # Discard old messages based on expiry
+            await connection.zremrangebyscore(
+                channel_key, min=0, max=int(time.time()) - int(self.expiry)
+            )
+
             # Check the length of the list before send
             # This can allow the list to leak slightly over capacity, but that's fine.
-            if await connection.llen(channel_key) >= self.get_capacity(channel):
+            if await connection.zcount(channel_key) >= self.get_capacity(channel):
                 raise ChannelFull()
+
             # Push onto the list then set it to expire in case it's not consumed
-            await connection.lpush(channel_key, self.serialize(message))
+            await connection.zadd(channel_key, time.time(), self.serialize(message))
             await connection.expire(channel_key, int(self.expiry))
 
     def _backup_channel_name(self, channel):
@@ -323,9 +330,9 @@ class RedisChannelLayer(BaseChannelLayer):
         # of the main message queue in the proper order; BRPOP must *not* be called
         # because that would deadlock the server
         cleanup_script = """
-            local backed_up = redis.call('LRANGE', ARGV[2], 0, -1)
+            local backed_up = redis.call('ZRANGE', ARGV[2], 0, -1)
             for i = #backed_up, 1, -1 do
-                redis.call('LPUSH', ARGV[1], backed_up[i])
+                redis.call('ZADD', ARGV[1], backed_up[i])
             end
             redis.call('DEL', ARGV[2])
         """
@@ -335,7 +342,15 @@ class RedisChannelLayer(BaseChannelLayer):
             # and the script executes atomically...
             await connection.eval(cleanup_script, keys=[], args=[channel, backup_queue])
             # ...and it doesn't matter here either, the message will be safe in the backup.
-            return await connection.brpoplpush(channel, backup_queue, timeout=timeout)
+            result = await connection.bzpopmin(channel, timeout=timeout)
+
+            if result is not None:
+                _, member, timestamp = result
+                await connection.zadd(backup_queue, float(timestamp), member)
+            else:
+                member = None
+
+            return member
 
     async def _clean_receive_backup(self, index, channel):
         """
@@ -343,7 +358,7 @@ class RedisChannelLayer(BaseChannelLayer):
         The result isn't interesting as it was already processed.
         """
         async with self.connection(index) as connection:
-            await connection.brpop(self._backup_channel_name(channel))
+            await connection.zpopmin(self._backup_channel_name(channel))
 
     async def receive(self, channel):
         """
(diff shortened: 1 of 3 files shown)


tarikki:
Another good-looking PR! :muscle: I'm not that versed in redis to write anything half as good :blush: 

1. Would it be possible to have some tests where we can capture the improvement brought on by this change?
2. Because the cleanup happens before `group_send` and `send` and the time complexity of ZREMRANGEBYSCORE is O(log(N)+M) (https://redis.io/commands/zremrangebyscore), the concern is that this might introduce latency when sending messages. The same concern was raised here regarding redis streams and expiring messages based on time: https://github.com/antirez/redis/issues/4450#issuecomment-345633165. I wonder if we should test against this to see at what `channel_capacity` values the latency becomes noticeable and mention this in the documentation?
3. Number 2 could be mitigated by having a flush command that is run regularly that would clean up all expired messages from all channels. Then the latency is shifted to a time when `channels_redis` is potentially idle instead of a time when `channels_redis` is guaranteed to be waiting to send messages.

astutejoe:
@tarikki You're absolutely right, I do feel that the expiration process for both is super janky and suboptimal. 
The best implementation I can think of is a separate process that ticks every second to clean expirations. That would also enable us to use simpler Redis data structures (such as lists again) and the cleanup process would only iterate over the expired messages and stop in the first unexpired message, yielding O(N) where N is the number of expired messages.

tarikki:
But all in all, I really like this implementation as it enforces expiration time per message. I think if most people are using the channels with the defaults (100 messages per queue / 60 sec expiration ) so the hit to performance on send should not be that big.

I think with the addition of: 

1. some improvements to documentation that describe this change.
2. a cleanup command that can be called periodically for power users.
3. some improvements to documentation on how to use the cleanup command.
4. tests that demonstrate single message expiration. (These should fail with the current system but work with this PR).

...this could be a good improvement to `channels_redis`. 

What do you think, @carltongibson? If this looks ok I'd be more than happy to work with @astutejoe with docs, tests and such.

carltongibson:
Hi @astutejoe @tarikki. Thanks for your work here: Sorry for the slow uptake: This pandemic has kind of knocked me sideways in terms of time.

I'd really like to pull together a new release here, with this and the other issues that got resolved recently. Currently the CI is broken, with a pytest version conflict. If either  of you have some capacity to look at that, and the other remaining issues/PRs I'd be grateful, we can rebase this, and we can get a release wrapped up. 

Thanks again!

tarikki:
@carltongibson: Sure. Would the CI problem go away by just updating the version of pytest, or do you think that would result in other conflicts?

 @astutejoe: Let's collab on this? If you could add me as editor to your fork of `channels_redis` I could start writing tests for the changes introduced in this PR and we could push this past the finish line together. I think I have time to work this next weekend.

carltongibson:
@tarikki Yes, I think, just updating the dependencies should be enough. Thanks!

astutejoe:
Done @tarikki

tarikki:
DONE

- Add tests to verify message expiration on `send()`

TODO

- Add tests to verify message expiration on `group_send()` 
- Run tests on different numbers of channels/groups/capacities to see when delay introduced by this feature becomes problematic
- Add function to flush all expired messages that can be called with celery etc. so keep the keep the queue clean instead of deleting messages just before send
- update documentation

@astutejoe: Could you please rebase on current master so tests will run? :blush:

astutejoe:
@tarikki rebased my fork and linted your tests with black.

tarikki:
@astutejoe: Thanks man. Need to setup black on my system as well.

chrisjbremner:
I think I have been running into a similar issue that this PR fixes. Is there anything else that needs to be done to push this over the finish line that I could help with?

astutejoe:
@tarikki do you think we need anything else to merge or is this PR not merged out of some miscommunication, disappearance-kinda issue?

tarikki:
I created a branch to test out the performance differences between this implementation and the current one.

I basically did some writes to channels with different capacities to see the difference between using a sorted set vs a list and created the branch to be able to do some asserts on how many messages a channel contains and also having the channel expiry time be greater than message expiry time (so the channel does not get deleted when the last message set to expire gets deleted). I ran the code on both the version in this PR and the version of channels_redis in staging. 

You can find the code here: https://github.com/astutejoe/channels_redis/tree/performance_tester. Wanna have a look to check that the logic is sound?

Here's the performance graphed. I think the most important section is the mass_push_average_time, which seems to be pretty consistent regardless of the channel capacities tested here (100 - 64000).  The strangest part of the graph is the  post_delete_one_message_push variation in performance times, but that is probably due to redis doing some internal cleaning.

image

The results seem indicate that using `channel_layer.send()` is twice as slow with a sorted set than using a list, but I think it could be an acceptable hit for having `channels_redis` work according to spec.

I created this branch ages ago and was planning on doing the same for `group_send`, but me and my wife bought an apartment three days after the work here and have been busy IRL :)

Full results: https://docs.google.com/spreadsheets/d/1CSg7MBOEXEApjhEauhMNhTiz6ELSXPXp_7FSRaQEFqs/edit?usp=sharing

channels_redis/core.py
```diff
@@ -628,25 +642,30 @@ async def group_send(self, group, message):
...
                         else
                             over_capacity = over_capacity + 1
                         end
                     end
                     return over_capacity
-                    """
-                % self.expiry
+                    """ % (
+                time.time(),
+                self.expiry,
```
carltongibson, line 668: Did this need adjusting per the comment on the issue? (Should we handle it separately?)

astutejoe: @carltongibson good point, will fix that later today

carltongibson:
@astutejoe I rebased and squashed.

channels_redis/core.py
carltongibson: @astutejoe Super thank. Throw it on top here as a separate commit (it's probably worth calling out separately.) If you fancy adding a line (or two) to CHANGELOG.txt that's cool too. (I can do it if not.) 

Thanks again!

astutejoe: @carltongibson There you go my dude, hope it's up to the standards. Feel free to change the CHANGELOG if the language isn't suitable or precise.

PR #185, opened 2020-03-06
Using UUID4 to guarantee uniqueness

### LINKED ISSUE #173: Channel names not unique
It seems channel names are generated using a global instance of random, which can cause problems if the seed is reset.  For instance see:
https://groups.google.com/forum/#!msg/otree/0UEtVVA_98M/9nUo0nLPAgAJ
Presumably, there is also a small risk of non-uniqure names even if the seed is not reset.

astutejoe:
From a quick benchmark, UUID4 generated 1 million unique strings in 4.04 seconds and the current implementation took ~7.4 seconds for the same. Fixes #173

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -5,11 +5,10 @@ import collections
 import hashlib
 import itertools
 import logging
-import random
-import string
 import sys
 import time
 import types
+import uuid
 
 import aioredis
 import msgpack
@@ -216,10 +215,7 @@ class RedisChannelLayer(BaseChannelLayer):
         self._receive_index_generator = itertools.cycle(range(len(self.hosts)))
         self._send_index_generator = itertools.cycle(range(len(self.hosts)))
         # Decide on a unique client prefix to use in ! sections
-        # TODO: ensure uniqueness better, e.g. Redis keys with SETNX
-        self.client_prefix = "".join(
-            random.choice(string.ascii_letters) for i in range(8)
-        )
+        self.client_prefix = uuid.uuid4().hex
         # Set up any encryption objects
         self._setup_encryption(symmetric_encryption_keys)
         # Number of coroutines trying to receive right now
@@ -533,12 +529,7 @@ class RedisChannelLayer(BaseChannelLayer):
         Returns a new channel name that can be used by something in our
         process as a specific channel.
         """
-        # TODO: Guarantee uniqueness better?
-        return "%s.%s!%s" % (
-            prefix,
-            self.client_prefix,
-            "".join(random.choice(string.ascii_letters) for i in range(12)),
-        )
+        return "%s.%s!%s" % (prefix, self.client_prefix, uuid.uuid4().hex,)
 
     ### Flush extension ###
 
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build/
 /.tox
 .DS_Store
 .pytest_cache
+.vscode
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -3,6 +3,8 @@ Unreleased
 
 * Updated msgpack requirement to `~=1.0`.
 
+* Ensured channel names are unique using UUIDs.
+
 
 2.4.2 (2020-02-19)
 ------------------
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1,4 +1,5 @@
 import asyncio
+import random
 
 import async_timeout
 import pytest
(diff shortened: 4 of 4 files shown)


channels_redis/core.py
```diff
@@ -209,10 +208,7 @@ def __init__(
         self._receive_index_generator = itertools.cycle(range(len(self.hosts)))
         self._send_index_generator = itertools.cycle(range(len(self.hosts)))
         # Decide on a unique client prefix to use in ! sections
-        # TODO: ensure uniqueness better, e.g. Redis keys with SETNX
-        self.client_prefix = "".join(
-            random.choice(string.ascii_letters) for i in range(8)
-        )
+        self.client_prefix = uuid.uuid4().hex
```
adamchainz: This is significantly more characters - could that be a problem?

astutejoe: @adamchainz I thought about it, but, the less characters, the less unique. I also thought about base32 or base64 to he more efficient.  Check the Redis Keys section:
 https://redis.io/topics/data-types-intro

astutejoe: @adamchainz Celery might be a good reference: https://github.com/celery/celery/blob/master/celery/app/base.py#L732

astutejoe: @adamchainz I'm not sure why they use the uuid from kombu but it's pretty much UUID4: https://kombu.readthedocs.io/en/latest/reference/kombu.utils.uuid.html

tarikki:
This looks pretty good and the following two tests fail with the old implementation but pass with this version:

```
@pytest.mark.asyncio
async def test_random_reset__channel_name(channel_layer):
    """
    Makes sure resetting random seed does not make us reuse channel names.
    """

    channel_layer = RedisChannelLayer()
    random.seed(1)
    channel_name_1 = await channel_layer.new_channel()
    random.seed(1)
    channel_name_2 = await channel_layer.new_channel()

    assert channel_name_1 != channel_name_2


@pytest.mark.asyncio
async def test_random_reset__client_prefix(channel_layer):
    """
    Makes sure resetting random seed does not make us reuse client_prefixes.
    """

    random.seed(1)
    channel_layer_1 = RedisChannelLayer()
    random.seed(1)
    channel_layer_2 = RedisChannelLayer()
    assert channel_layer_1.client_prefix != channel_layer_2.client_prefix

```


Also the length is probably not a problem because it's still under the maximum of 100 bytes mentioned in the Channel Layer Specification / Channel Semantics: https://channels.readthedocs.io/en/latest/channel_layer_spec.html

tarikki:
@astutejoe: Wanna add those two test to the PR? :blush:

astutejoe:
> @astutejoe: Wanna add those two test to the PR? 😊

Will do! :D

astutejoe:
@tarikki there you go!

channels_redis/core.py
adamchainz: This is good but UUID's actually have some predictability to them. Using the hex value is also less efficient as you noted. For true randomness, I think it'd be best to turn to `get_random_string` from inside Django: https://github.com/django/django/blob/3c35825009540e222311d591c7330fa17dd76c99/django/utils/crypto.py#L54 . You can use length 22 to get 131 bits randomness, which is more than the 128 bits that a UUID4 uses in 32 hex characters.

adamchainz:
Thanks for the tests @tarikki

channels_redis/core.py
carltongibson: I´m only following this from the side-lines 🤹‍♂ but surely UUIDs are random enough here? 

This is just to stop naming collisions right, rather than anything _secure_ that we need to stop baddies from guessing?

Do we need to do anything more?

astutejoe: @carltongibson Yeah, it's definitely not cryptographically secure, but I really don't see any reasons we should ensure cryptographically secure randomness, with access to Redis an attacker could most probably do anything, randomized keys or not. Bear in mind that cryptographically secure randomization can be considerably more expensive.

astutejoe: @carltongibson To be more specific, the implementation in my PR is cryptographically secure but only for CPython: https://stackoverflow.com/a/53542384/1418283

adamchainz: You're right, we don't need crypto grade security. `get_random_string` is still denser and so wins some storage space but that may not be a concern. Happy to merge as-is.

mpasternak:
@astutejoe , @tarikki , thanks for your work! I came here to submit this exact fix, I'm glad to see all this work already done. @adamchainz , you're lucky guy having all this help from the community. If there's anything I could help you all with to get this issue merged, please let me know.

mpasternak:
@carltongibson  THANK YOU THANK YOU THANK YOU

PR #188, opened 2020-04-10
Add more check when returning a connection to connection pool

tinylambda:
If call push method with a connection that created by the pool object, it should not raise a KeyError, so I add some check in the implementation.


carltongibson:

PR #199, opened 2020-07-01
Lowered group_send failure logging level from exception to info.

### LINKED ISSUE #181: "channels over capacity in group xxx" flood
Since 2.42, we're getting tons of "channels over capacity in group xxx" messages, even though we make little use of django channels (just a couple admin users).
Still haven't found out exactly why, but the fact is: throwing this message in every`RedisChannelLayer.group_send()` call can be *extremely* noisy.
It would be nice to find a way to throttle this!

carltongibson:
Closes #181.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -7,7 +7,9 @@ Unreleased
 
 * Ensured messages are expired even when channel is in constant activity.
 
-* Optimized Redis script caching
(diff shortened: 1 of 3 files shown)


ramwin:
I consider it was a bad idea to lower the logging level. It made the error invisiable to the developer. I recommend use error by default and make a config option for the developer.

PR #205, opened 2020-07-07
Update 3.0.0 changelog entry to note redis version

driverdan:
3.0.0 introduces a breaking change requiring redis >= 5.0. This should be noted so anyone reviewing changes will know it's a breaking change. The README was updated but the changelog was not.


carltongibson:
Ok, we can add this, but only the current version and the previous one are supported according Redis’ policy, so it amounts to “Don’t use an EOL version of Redis”, which (surely) isn’t a change...?

I’d like to like back to the Redis docs here, and add a similar note to the README. Would you be able to add that? (I can do that if not)

Issue #212, opened 2020-07-24
Slow memory leak in RedisChannelLayer .receive_buffer due to lack of cleanup/expiry


ryanpetrello:
I'm observing what looks like a slow memory leak in `channels-redis` in a project I maintain that uses Daphne (Ansible AWX):

```
/var/lib/awx/venv/awx/bin/pip3 freeze | egrep "chan|daphne|Twisted"
channels==2.4.0
channels-redis==2.4.2
daphne==2.4.1
Twisted==20.3.0
```

https://github.com/ansible/awx/blob/devel/awx/main/consumers.py#L114

This looks _quite_ similar to a recently reported (and resolved) memory leak: https://github.com/django/channels/issues/1181

After noticing a slow memory leak in Daphe at high volumes of (large) messages:

![image](https://user-images.githubusercontent.com/214912/88402358-02ddc480-cd99-11ea-9047-9d825ca1496d.png)

I decided to add some instrumentation to my routing code with `tracemalloc` to try to spot the leak:

```diff
diff --git a/awx/main/routing.py b/awx/main/routing.py
index 2866d46ed0..71e40c693f 100644
--- a/awx/main/routing.py
+++ b/awx/main/routing.py
@@ -1,3 +1,6 @@
+from datetime import datetime
+import tracemalloc
+
 import redis
 import logging

@@ -10,11 +13,16 @@ from channels.routing import ProtocolTypeRouter, URLRouter
 from . import consumers


+TRACE_INTERVAL = 60 * 5  # take a tracemalloc snapshot every 5 minutes
+
+
 logger = logging.getLogger('awx.main.routing')


 class AWXProtocolTypeRouter(ProtocolTypeRouter):
     def __init__(self, *args, **kwargs):
+        tracemalloc.start()


+class SnapshottedEventConsumer(consumers.EventConsumer):
+
+    last_snapshot = datetime.min
+
+    async def connect(self):
+        if (datetime.now() - SnapshottedEventConsumer.last_snapshot).total_seconds() > TRACE_INTERVAL:
+            snapshot = tracemalloc.take_snapshot()
+            top_stats = snapshot.statistics('lineno')
+            top_stats = [
+                stat for stat in top_stats[:25]
+                if 'importlib._bootstrap_external' not in str(stat)
+            ]
+            for stat in top_stats[:25]:
+                logger.error('[TRACE] ' + str(stat))
+            SnapshottedEventConsumer.last_snapshot = datetime.now()
+        await super().connect()
+
[...]

PR #213, opened 2020-07-27
implement a TTL for RedisChannelLayer.receive_buffer (to avoid a memory leak)

ryanpetrello:
see: https://github.com/django/channels_redis/issues/212

cc @carltongibson @agronick

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -179,6 +179,40 @@ class UnsupportedRedis(Exception):
     pass
 
 
+class ExpiringCache(collections.defaultdict):
+    def __init__(self, default, ttl=60, *args, **kw):
+        collections.defaultdict.__init__(self, default)
+        self._expires = collections.OrderedDict()
+        self.ttl = ttl
+
+    def __setitem__(self, k, v):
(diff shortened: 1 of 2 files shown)


channels_redis/core.py
```diff
@@ -616,6 +651,7 @@ async def group_discard(self, group, channel):
         key = self._group_key(group)
         async with self.connection(self.consistent_hash(group)) as connection:
             await connection.zrem(key, channel)
+        self.receive_buffer.expire()
```
ryanpetrello, line 654: Maybe we can't rely on this *only* happening on `group_discard` (since some may not be using the group extension?).

Doing this sort of thing once per `receive()` call seems too expensive, though.

ryanpetrello:
Using https://github.com/ryanpetrello/channels-redis-leak as a reproducer, I've applied this patch.

Here's a bunch of references to messages:

image

...here's what happens if I wait 60+ seconds and disconnect/reconnect w/ a new websocket client:

image

ryanpetrello:
On further testing, this seem like it also slowly leaks 😞

PR #219, opened 2020-08-20
respect channel layer capacity in RedisChannelLayer.receive_buffer

ryanpetrello:
cc @carltongibson 

respect the capacity setting so that the receive_buffer does not grow
without bounds (which can lead to a memory leak)

see: https://github.com/django/channels_redis/issues/212

based on the idea outlined here:

https://github.com/django/channels/issues/1361#issuecomment-672926804

> I'd argue that these per-channel Queue objects should probably be bound in size. There's already a capacity argument; maybe the per-channel buffer should respect that, and only buffer up to that many objects before dropping old ones?

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -2,6 +2,7 @@ import asyncio
 import base64
 import binascii
 import collections
+import functools
 import hashlib
 import itertools
 import logging
@@ -179,6 +180,19 @@ class UnsupportedRedis(Exception):
     pass
 
 
+class BoundedQueue(asyncio.Queue):
+    def put_nowait(self, item):
+        if self.full():
+            # see: https://github.com/django/channels_redis/issues/212
+            # if we actually get into this code block, it likely means that
+            # this specific consumer has stopped reading
+            # if we get into this code block, it's better to drop messages
+            # that exceed the channel layer capacity than to continue to
+            # malloc() forever
+            self.get_nowait()
+        return super(BoundedQueue, self).put_nowait(item)
+
+
 class RedisChannelLayer(BaseChannelLayer):
     """
     Redis channel layer.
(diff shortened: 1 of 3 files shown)


ryanpetrello:
Hey @carltongibson,

Any chance you've had some time to look over this PR?  I have a number of AWX users that are using a version of this patch, and it looks promising (none have yet reported back the memory leak after applying this patch).

carltongibson:
Hi @ryanpetrello. I'm just back from holiday today (and had three versions of Django out this morning). 

I'll try and give it a look this week. Thanks for the hustle! 😀

ryanpetrello:
Awesome - thanks!

carltongibson (commented):
Hi @ryanpetrello — I know this is a slow leak but do you think you can add some basic regression tests of the the new queue?

ryanpetrello:
Hey @carltongibson - what sort tests are you looking for?  Something that verifies that the queue won't grow beyond the configured capacity?

carltongibson:
Yeah, that sounds fine. "Did the basic behaviour break?" is enough.

ryanpetrello:
@carltongibson test added - let me know if this is what you're looking for.

carltongibson:
@ryanpetrello Quick glance, looks about right. Thank you. 

I'll have one more look over the issues and then we'll bundle this up for a release. 👍

Thanks for your input here! 🥇

channels_redis/core.py
```diff
@@ -544,7 +560,11 @@ async def new_channel(self, prefix="specific"):
         Returns a new channel name that can be used by something in our
         process as a specific channel.
         """
-        return "%s.%s!%s" % (prefix, self.client_prefix, uuid.uuid4().hex,)
+        return "%s.%s!%s" % (
```
ryanpetrello, line 563: @carltongibson I changed this because `black` was complaining about it

carltongibson: No problem. Whatever black says. 😀

PR #220, opened 2020-09-03
consistent_hash parameter in the send method is incorrect

weibao:


carltongibson (commented):
Hi @weibao — we'll need to add a test here. Just trying to think of the simplest thing that's sufficient. 🤔

carltongibson:
@weibao I rebased to fix the lint error. 🤞

PR #224, opened 2020-09-11
add special symbol password usage in document

### LINKED ISSUE #143: how can i auth redis with a special symbol in password
```
CHANNEL_LAYERS = {
    "default": {
        "BACKEND": "channels_redis.core.RedisChannelLayer",
        "CONFIG": {
            "hosts": ["redis://:xxx#xxxxx@ip:port/8"],
        },
    },
}
```
eg: my password has a “#” in it.
i tried replace it to "%23"
i tried urllib.parse.quote and quote_plus
all failed

betta-cyber:
to add special symbol password usage in document. 
Fixes #143

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -43,6 +43,8 @@ but note that changing the host list will lose some sharded data.
 If your server is listening on a UNIX domain socket, you can also use that to connect: ``["unix:///path/to/redis.sock"]``.
 This should be slightly faster than a loopback TCP connection.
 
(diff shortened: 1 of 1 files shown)


KelsamGames:
Thank you for this! This (or another info-equivalent update) should definitely be merged.

carltongibson (commented):
I don't really think this is appropriate for the docs here. It's the kind of addition that belongs in the docs for the underlying Redis library, which handles the connection. (All `channels_redis` does is hand the connection params down, to in this case the linked `create_connection`)

Issue #226, opened 2020-09-16
Add a list of redis commands (documentation improvement)


bastbnl:
I stumbled on [this post](https://www.digitalocean.com/community/tutorials/how-to-install-and-secure-redis-on-debian-9) and decided to apply the `rename-command` documented in step 5 - Renaming Dangerous Commands on the redis server used by the project I'm working on. This project has `celery`, `channels_redis` and `django-redis`, so I thought I'd take a look to see which commands could be renamed.

Looking through the celery RedisBackend, I found that this project [includes a list of celery commands](https://github.com/celery/celery/blob/v4.4.7/celery/backends/redis.py#L187) used by the backend and I was unable to find such a list for channels_redis, so I decided to add one in a PR. I'm not sure that this list is the complete list of commands used by the modules, but it should contain the most obvious commands

PR #227, opened 2020-09-16
Added a list of redis commands, much like the celery RedisBackend

bastbnl:
This is intended to "fix" issue #226. I've taken some time to find the obvious redis commands, but I'm not sure if this is the complete list.


vanschelven:
#291 is a (currently) more complete list.

PR #228, opened 2020-10-29
Bumped dependencies to allow updating.

carltongibson:
asgiref to match Django 3.1
Channels to <4 pending release.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,63 @@
+name: Tests
+
+on:
+  push:
+#    branches:
+#    - master
```
carltongibson, line 6: Commented for now, since it didn’t run otherwise.

PR #237, opened 2021-02-09
Use DjangoJsonEncoder in message serialization

sevdog:
When using django developers are used have some standard-library types JSON serialized correctly (thanks to `DjangoJSONEncoder`).

### Problem
Previous to this PR running this code

```python
import uuid
from channels_redis.core import RedisChannelLayer
layer = RedisChannelLayer()
message = {'uuid': uuid.uuid4()}
layer.serialize(message)
```

Would raise this error

```
msgpack/_packer.pyx in msgpack._cmsgpack.Packer._pack()
TypeError: can not serialize 'UUID' object
```

### Solution

`msgpack` packer has an undocumented (on RTD but documented in source) param [`default`](https://github.com/msgpack/msgpack-python/blob/cfae52437b0d146f74a14d69cb706712c1fa2c95/msgpack/fallback.py#L672-L674) which acts similar to [`json.dump` omonimous param](https://docs.python.org/3/library/json.html#json.dump).

This PR adds support for `DjangoJsonEncoder` in `RedisChannelLayer.serialize` method by providing `msgpack.packb` with an appropriate default callback.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -15,6 +15,7 @@ import uuid
 import aioredis
 import msgpack
 
+from django.core.serializers.json import DjangoJSONEncoder
 from channels.exceptions import ChannelFull
 from channels.layers import BaseChannelLayer
 
@@ -248,6 +249,8 @@ class RedisChannelLayer(BaseChannelLayer):
         # Per-channel cleanup locks to prevent a receive starting and moving
         # a message back into the main queue before its cleanup has completed
         self.receive_clean_locks = ChannelLock()
+        # JSON encoder for non-basic-types
+        self._encoder = DjangoJSONEncoder()
 
     def decode_hosts(self, hosts):
         """
@@ -811,7 +814,7 @@ class RedisChannelLayer(BaseChannelLayer):
         """
         Serializes message to a byte string.
         """
-        value = msgpack.packb(message, use_bin_type=True)
+        value = msgpack.packb(message, use_bin_type=True, default=self._encoder.default)
         if self.crypter:
             value = self.crypter.encrypt(value)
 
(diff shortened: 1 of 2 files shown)


carltongibson (changes requested):
Hi @sevdog — thanks for the suggestion here. 

I don't think bringing in the dependency on Django here is correct. 

We **could** add a config option I suppose. So users would optionally specify their preferred encoder, but what happens on decoding? You're not going to get back the types you passed in. 

For that reason I'm inclined to think this is out of scope for the channel layer. Map from/to serializable dicts in your client code before/after sending/receiving from the channel layer. 🤔

sevdog:
Hi @carltongibson,

> I don't think bringing in the dependency on Django here is correct.

IMO there should not be any problem with django-dependency, in the end `channels_redis` is a plugin for `channels` which is designed around `django`. If you install this library you **also need** to install django. I do not see why this should be hidden in the core of this package (see dependency tree).

```
channels-redis
  - channels [required: <4]
    - Django [required: >=2.2]
```

> We could add a config option I suppose. So users would optionally specify their preferred encoder, but what happens on decoding? You're not going to get back the types you passed in.

This is a good idea, it could be set in `CHANNEL_LAYERS['<layer>']['CONFIG']` as a dotted path. However to have the real method to be used I would use [`import_string`](https://docs.djangoproject.com/en/3.1/ref/utils/#django.utils.module_loading.import_string) from django utils (has channels does for `CHANNEL_LAYERS['<layer>']['BACKEND']`), because importlib as of now does not support importing a single element from module. I would like avoiding to rewrite that method since it will not be DRY.

Regarding de-serialized content is obvious that types will change (ie: complex types becomse strings), IMO this should be documented (even if this should obvious when you serialize complex elemens) and be left to users to handle de-serialization.

Let me know what you think and then I will update this PR.

PR #239, opened 2021-04-07
aioredis sentinel

qeternity:
Hi all - so after coming back to this, I've noticed that aioredis now has sentinel support natively which is an easier change than porting over to a completely new library (as with my now broken aredis branch).

We have deployed this to our testing environment and will stress test it over the coming days.

I've not included any unit tests here so feedback on how this should be accomplished is appreciated.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -48,8 +48,9 @@ class ConnectionPool:
     taking into account asyncio event loops.
     """
 
-    def __init__(self, host):
+    def __init__(self, host, master_name=None):
         self.host = host
+        self.master_name = master_name
         self.conn_map = {}
         self.in_use = {}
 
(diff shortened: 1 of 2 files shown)


carltongibson:
Hey @qeternity — thanks for this. So aioredis is alive again? (I haven't looked into it.)

qeternity:
It appears it is back alive again, and planning a 2.0 release. The latest 1.3.x branch which is over a year old has sentinel support, and with the apparent resumption of activity, it no longer makes sense for us to maintain an aredis fork and I'd rather upstream sentinel support.

This current PR is very much alpha - but I'd like to get the ball rolling. It's about time that channels can be deployed in an HA environment.

carltongibson:
OK, super. 💃 

I guess the first question is what's happened to Python 3.7? 😀

qeternity:
Yeah interesting one that is. I'll have to dig in.

qeternity:
I am closing this. We are unfortunately abandoning our efforts with this. The aioredis library has some nasty quirks when working with sentinels. This introduces some tail latencies due to sync blocking under the hood (crazy for a library named aioredis).

I think anyone seeking HA channels will need to use KeyDB multi master or use the competing rabbitmq layer.

carltongibson:
Hey @qeternity — OK, thanks for the follow-up. 

Just so I'm clear, your conclusion would be that actually moving to `aredis` (or some other option) would be the way forwards here?

qeternity:
I think reviewing the soon to be released 2.0 is probably the next order of business. But unless they have resolved some of the issues then yes I would say aredis is probably the next best thing.

I would also say that using the native connection pool in whichever library would probably help smooth out some of the issues we experienced.

PR #240, opened 2021-04-08
Don't declare wheels as universal

michael-k:
That's intended for wheels that support both Python 2 and 3.


carltongibson (approved):
Yep, super.

PR #241, opened 2021-04-10
remove strict_map_key requirement for deserialize

envirocoder:
Trying to send a dict with integer keys, 
e.g.
```
async_to_sync(channel_layer.group_send)(
        "name",
        {"type":"test",
        "data": {1:10, 2:20},
        }
```
I got 
```
  File "msgpack/_unpacker.pyx", line 195, in msgpack._cmsgpack.unpackb
ValueError: int is not allowed for map key
```
not familiar with the channels or redis code to know if there will be implementations else where; is it safe to relax the checking in  msgpack unpacking with strict_map_key=False ?


envirocoder:
just noticed, duplicate of #233

PR #242, opened 2021-04-15
pipeline message garbage collection

qeternity:
With large groups, we end up awaiting the garbage collection on each member channel. This PR moves that into a redis pipeline so we only await a single call.


carltongibson (approved):
Yep, that makes sense. Thanks!

PR #244, opened 2021-04-25
RFC - use aioredis pools

qeternity:
Hi all - had some time this weekend to re-evaluate getting sentinel support in here after a false start a couple weeks back. After going through aioredis in more detail, it struck me that the easiest approach forward is to hand off the pool orchestration to the native pool handlers.

I'm far from an expert in either of these libraries, so this is more of an RFC than a concrete PR. But it strikes me that having a native pool per loop and letting the underlying library manage the rest is probably the cleanest approach. It also allows sentinel compatibility without having to manage connection/sentinel relationships.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -120,8 +120,9 @@ class ConnectionPool:
                 self.sentinel_map[conn].close()
                 await self.sentinel_map[conn].wait_closed()
                 del self.sentinel_map[conn]
-            conn.close()
-            await conn.wait_closed()
+            else:
+                conn.close()
+                await conn.wait_closed()
             del self.conn_map[loop]
 
     async def close(self):
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -48,96 +48,95 @@ class ConnectionPool:
     taking into account asyncio event loops.
     """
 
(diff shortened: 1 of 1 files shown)


qeternity:
I just refactored to the old naming convention to make the PR a bit easier to understand.

Obviously there are quite a few tests that are failing due to all sorts of checks that aren't needed anymore, and the different shape of the `conn_map` which is now a single aioredis pool as opposed to a list of connections.

I am happy to push forward with this if others think it's a good idea, but before that I wanted to poll the audience.

carltongibson:
Hey @qeternity — just a quick look, but this seems like a good idea. If we can hand off connection management then that's a win no?

qeternity:
To split this up in something that is a bit more manageable, we are now running a branch in our test env that uses the native aioredis pools but keeps the current connection pooling overlay (a bit less efficient given that each connection is now a pool under the hood). I've also added a redis sentinel service to test and duplicated the existing tests to run entirely under sentinels.

I see a few more things which need doing:

1. unit tests around sentinel failure/degradation scenarios
2. better sentinel configuration options (we don't use multiple redis db)

https://github.com/zumalabs/channels_redis/pull/3

Issue #229, opened 2020-10-29
Intermittent ChannelFull test failure


carltongibson:
e.g. https://github.com/django/channels_redis/runs/1328351504?check_suite_focus=true#step:6:105

This is an artefact: not consistent on run or version. 

Need to make the test more reliable.

carltongibson:
Going to note which test(s) fail(s). (If it's always the same one...)

https://github.com/django/channels_redis/runs/2919887893?check_suite_focus=true
```
FAILED tests/test_core.py::test_receive_cancel
```

carltongibson:
https://github.com/django/channels_redis/pull/262/checks?check_run_id=3016793723

```
FAILED tests/test_core.py::test_receive_cancel - channels.exceptions.ChannelFull
```

carltongibson:
https://github.com/django/channels_redis/pull/262/checks?check_run_id=3017259693

```
FAILED tests/test_core.py::test_receive_cancel - channels.exceptions.ChannelFull
```


Looks like doubling the capacity in `test_receive_cancel` might be required here.

carltongibson:
```
tests/test_core.py::test_receive_cancel - channels.exceptions.ChannelFull
```

https://github.com/django/channels_redis/pull/274/checks?check_run_id=3492323345

Ok, it's looks like it's always the same test. 

Just bumping the limit is probably worth a go.

revoteon:
I went ahead and did some debugging in brute force way. I removed `ChannelFull` exception and instead printed out values of capacity and current count to see how far it can overflow with the tests above. Then I ran the test 100 times. The count never exceeded 15.  

I bumped capacity to of channel layer to 20 in `test_receive_cancel` - both in `test_core.py` and in `test_sentinel.py`. Then I ran tests again 100 times. I didn't get any exception.

So I think your suspicion is correct; just bumping the limit twice to 20 seem to resolve issue for both of these tests.

carltongibson:
Super @revoteon. Want to do a PR and we'll get that in. 

Thanks for the effort! 🎁

revoteon:
@carltongibson Done! #277

PR #246, opened 2021-04-28
redis sentinel support

qeternity:
To capitalize on the bit of interest this library seems to have at the moment, I have pushed the internal branch that we have been using that provides limited sentinel support, and switches over to using aioredis pools (necessary as part of the sentinel client).

Even if this doesn't get merged because a pubsub rewrite supersedes it, it might have some value to others. A few things to note:

1. ~~There is no database support when using sentinels. This would be an easy add-on but we don't use internally, instead opting to scale out the number of redis clusters. But it doesn't fit in with the existing config shape.~~
2. ~~There is no sharding support when using sentinels. The config expects a list of tuples corresponding to the sentinel cluster. Again, an easy tweak but not something we use internally.~~
3. The tests have been duplicated from the vanilla redis tests, and don't test any specific sentinel functionality around failover. But in a failover situation, the connections get reaped by the aioredis pool, which triggers other connection cleanup so not entirely sure if that is out of scope for the unit tests.
4. The test suite only provides one sentinel and one redis instance.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,6 @@ jobs:
         - 3.7
         - 3.8
         - 3.9
-
     services:
       redis:
         image: redis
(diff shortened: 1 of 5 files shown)


qeternity:
We also witness the intermittent test failure from https://github.com/django/channels_redis/issues/229

qeternity:
@carltongibson having looked at the PubSub implementation, I realized that it's going to basically be a layer rewrite which still leaves the `ConnectionPool` intact, so a PR (or probably two) that add sentinel support and moves to using the native connection pooling is probably still in order.

qeternity:
Ok @carltongibson so I've gone back and tweaked the shape of the settings so that we're at feature parity now with vanilla redis setups (fixed issues 1 and 2 from the above list). It would be great to get this merged, and I can return focus to using native aioredis pools which would be a nice to have hand in hand with the PubSub layer.

qeternity:
@carltongibson what say you?

carltongibson:
Hey @qeternity — "Yes!" 😀

I just need half a cycle to look at it all. More or less "Merge all the things" and do a release is the plan.

carltongibson:
Moving to #250.

PR #247, opened 2021-04-28
Provide an alternative channel layer implementation that uses Redis Pub/Sub

acu192:
Fixes Issue #245


acu192:
There is one logic bug, but I'll propose that we **do not** fix it.

This implementation does not support a Consumer doing this:
```python
self.channel_layer.group_add("whatever_group", someone_elses_channel_name)
```

Rather, you need to stick with the normal pattern of only adding *yourself* to groups:
```python
self.channel_layer.group_add("whatever_group", self.channel_name)
```

Stated another way: One consumer cannot subscribe a *different* consumer (denoted by `someone_elses_channel_name` above) to a group.

Again, I would propose that we let this be a documented limitation of this implementation. Fixing it is possible by publishing special messages destined for `someone_elses_channel_name` to tell them to do their own subscribing... but that logic will be convoluted, error-prone, and I doubt this will actually be an issue for most users anyway, so not worth it.

acu192:
The other thing to note: Not all configuration parameters are supported (yet).

**expiry**

I don't think this is needed. A consumer's own message queue will be cleaned up when that consumer disconnects, which essentially "expires" those messages for us (removes them from memory, that is).

**group_expiry**

Similarly, a group naturally "expires" when no one is subscribed to it.

**capacity**

This might make sense to implement. It wouldn't be hard, just add a check of the message queue length inside `_do_receiving()`.

**channel_capacity**

Same...

**symmetric_encryption_keys**

This would be easy enough to add, just pass these keys along to the two places where `aioredis.create_redis()` is used.

aryan340:
@acu192 but there is a workaround for this.
In order to achieve this,
each user **must** have a private channel of their own.

When you need to add an user to a group, you send them a special message internally.
```py
await self.channel_layer.group_send('user_private_channel', 'user_channel_name', {
    'type': 'join_group',
    'group_name': 'group_name_here'
    }
)
```
Similarly, for leaving groups, we could have an event type called `leave_group`.
When the consumer receives this special event, it must call `self.channel_layer.group_add` internally.
These event types must be reserved and hence we can actually add channels to groups from outside.

If you have any more questions or I haven't explained this clearly. please let me know.
thanks!

acu192:
> each user must have a private channel of their own

This is already the case (via `self.channel_name` which is created by the channel layer's `new_channel()` method). But maybe I'm not understanding what your proposing; are you proposing that each consumer has a *second* private channel?

> If you have any more questions or I haven't explained this clearly. please let me know.

I'm interested in thinking about this more, although I'm not seeing this as a solution yet. For example, a common use-case in my app (and I assume other apps) is to add a channel to a group then immediately do a `group_send()` to that group. In such a case, that new channel should get that group message (because your `await group_add(...)` returned, so you would assume it *happened*). However, with the solution you propose there would be a delay between the `await group_add(...)` returning to when the channel is **actually added** to the group (because of those special messages propagating through and needing to be processed on the other end). Therefore, a `group_add()` followed immediately by a `group_send()` would work if the channel was in-process, but would not work if the channel was "outside" the process (to use your term).

One idea, trying to solve it, would be for the caller of `group_add()` to wait for an acknowledgement message from the other consumer, saying "got it. I added myself like you told me to". That way, the caller of `await group_add(...)` would know that when it returns, it *happened*. But... here is where the real complexity comes in. What if you never get an ack? Entirely possible because the other consumer might have exited already. So maybe we have a timeout? This is where it begins to feel extremely error-prone.

@aryan340 Out of curiosity, do you need this functionality in your app? I.e. adding channels to groups from the "outside"?

acu192:
I just added a runtime check to `raise` on the known limitation of this implementation (i.e. not being able to do a `group_add(...)` on channels that are not in-process).

I propose we merge this PR with this known limitation. It is a low-risk merge (it does not change any existing code) and we can continue improving it over time. @aryan340 Perhaps after this PR is merged, you can send in a new PR with the workaround you proposed?

Maintainers, what do you think?

aryan340:
>@aryan340 Out of curiosity, do you need this functionality in your app? I.e. adding channels to groups from the "outside"?

Yes, I would be needing that. My use case is, users are a part of boxes. Whenever they join a box, I listen to that via a listener and add the user to the associated group. So this needs to be done outside the consumer.

>This is already the case (via self.channel_name which is created by the channel layer's new_channel() method). But maybe I'm not understanding what your proposing; are you proposing that each consumer has a second private channel?

Sorry I meant that each consumer must have it's own private group.

aryan340:
>One idea, trying to solve it, would be for the caller of group_add() to wait for an acknowledgement message from the other consumer, saying "got it. I added myself like you told me to". That way, the caller of await group_add(...) would know that when it returns, it happened. But... here is where the real complexity comes in. What if you never get an ack? Entirely possible because the other consumer might have exited already. So maybe we have a timeout? This is where it begins to feel extremely error-prone

You can configure redis pub sub clients to listen for subscribe or unsubscribe messages. Everytime a client subscribes a message is sent. This way we can know if the consumer is finished subscribing.

aryan340:
@acu192 I am okay if this PR doesnt provide external group add support out of the box.
Based on my idea, anyone can just write handlers for the above use case, and its not a big deal.
But if you insist, I can make a PR too.

acu192:
> But if you insist, I can make a PR too.

No, I'm not insisting; it would be *totally up to you* if you wanted to try to extend this implementation with your idea.

But first we need the maintainers' buy-in to *this* PR.

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+            conn = await self._get_sub_conn()
+            await conn.subscribe(self._receiver.channel(channel))
+
+    async def unsubscribe(self, channel):
+        if channel in self._subscribed_to:
+            self._subscribed_to.remove(channel)
+            conn = await self._get_sub_conn()
+            await conn.unsubscribe(channel)
+
+    async def _get_pub_conn(self):
```
aryan340, line 201: I think that instead of getting a connection each time, each consumer should get their own pubsub client class.
And also, I think that there is no need of keeping track of subscribed channels locally.

Have a redis.pubsub client initialized and then get the channels subscribed by doing client.channels

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+    unsubscribes. That is not what we want; instead, we want the Receiver
+    to continue even if no one is subscribed, because soon someone *will*
+    subscribe and we want things to continue from there. Passing this
+    empty function solves it.
+    """
+    pass
+
+
+class RedisSingleShardConnection:
+    def __init__(self, host, channel_layer):
```
aryan340, line 174: initialize a redis.pubsub instance here.
We will be using the create_redis instance only for publishing.
use the internal client for subscribing.

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+        assert (
+            isinstance(hosts, list) and len(hosts) > 0
+        ), "`hosts` must be a list with at least one Redis server"
+
+        self.prefix = prefix
+
+        # Each consumer gets its own *specific* channel, created with the `new_channel()` method.
+        # This dict maps `channel_name` to a queue of messages for that channel.
+        self.channels = {}
+
```
aryan340, line 25: I don't understand why this channel layer keeps a local copy of groups and the channels subscribed to groups. If this is the case, then this implementation won't scale and I think it won't be any different than the previous one.

Rather, I would suggest just let redis handle this natively and make calls from the internal client whenever we need to know which channels are subscribed.

Also, I mean that we should not keep track of clients in sets from the channel layer.
instead we need to initialize the client from the consumer class.

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+        self._receiver = None
+        self._receive_task = None
+        self._keepalive_task = None
+
+    async def publish(self, channel, message):
+        conn = await self._get_pub_conn()
+        await conn.publish(channel, msgpack.packb(message))
+
+    async def subscribe(self, channel):
+        if channel not in self._subscribed_to:
```
aryan340, line 190: the native redis client handles whether we are connected already, doesn't it?
I think that it is better to just call subscribe and let redis handle it at the back natively.

channels_redis/alt.py
acu192: > I think that instead of getting a connection each time, each consumer should get their own pubsub client class.

So there would be as many connections to Redis as there are consumers? I think that is excessive, and probably less efficient.

> Have a redis.pubsub client initialized and then get the channels subscribed by doing client.channels

We need to keep our own set of channels so that we can resubscribe if we lose connection to Redis. `aioredis` doesn't reconnect nor resubscribe on its own, so we have to do that ourselves.

channels_redis/alt.py
acu192: Well, we cannot connect to redis from `__init__` because we need a coroutine (`__init__` is synchronous; not a coroutine).

> use the internal client for subscribing

Not sure what you mean. You cannot publish and subscribe/listen from the same redis connection. Rather you need one connection for publishing, and another for subscribing/listening.

channels_redis/alt.py
acu192: You might have more than one consumer that is subscribed to a single group ("channel" in redis terminology). You can only subscribe to that channel once per redis connection. Think of it this way: Redis publishes inter-process for us (i.e. Redis does [IPC](https://en.wikipedia.org/wiki/Inter-process_communication)), and we do our own publishing intra-process (being intra-process, it's as fast as it can be...because there's no blocking or anything because we are intra-process). So the way we are doing it here is actually fastest.

channels_redis/alt.py
aryan340: >So there would be as many connections to Redis as there are consumers? I think that is excessive, and probably less efficient.

there would be as many connections, but we don't need to keep track of a set of connected clients.
Besides, this is the way to go. You need to have concurrent clients connected to redis and then do centralized publishes. Otherwise, this is no different from the current channel layer - where we have one single connection and use redis as a "message queue". 
We need to be broadcasting messages instead.

channels_redis/alt.py
acu192: I don't see a reason to do a whole network round-trip to-and-from Redis to subscribe to a channel we are already subscribed to. Maybe `aioredis` short-circuits this for us in this case (to avoid the network round-trip call to Redis), but I don't know.

Keep in mind with `asyncio` such checks don't risk race conditions because context switching only happens at `await`s. It's fast and safe to do our own check here.

channels_redis/alt.py
aryan340: >Not sure what you mean. You cannot publish and subscribe/listen from the same redis connection. Rather you need one connection for publishing, and another for subscribing/listening.

I get you. What I meant was, we have one single connection for pulishing.
Each consumer has their own connection which they use to subscribe.

channels_redis/alt.py
aryan340: >We need to keep our own set of channels so that we can resubscribe if we lose connection to Redis. aioredis doesn't reconnect nor resubscribe on its own, so we have to do that ourselves.

When we have separate connections per consumer,
why not detect if we lost connection from the consumer and reconnect individually?
this way we don't need to keep track of the clients in a set.

channels_redis/alt.py
aryan340: but then we have a tradeoff here.
This implementation cannot hold more than a 1000 channels as a whole, because we locally store these in sets.
In order for this implementation to be scalable we need to avoid doing so - even if it means more calls to redis.

aryan340: @acu192 
the whole point of using pub sub is to utilize the scalability redis gives. Having local caches completely destroys this purpose. Secondly, we don't have connections per consumer. We are not even using the pub sub feature properly here.
This implementation is only having two connections - one to listen and one to subscribe. In other words, it is keeping clients in a local set, then keeping track of what channels they are subscribed to, and subscribing for them via the second connection. So for a lot of consumers this won't work out - if you see what I mean. Instead, it is better to have a connection per consumer and then let redis handle all the subscriptions, natively. This would be the best implementation possible.
what do you say @carltongibson ?

carltongibson:
Hey @acu192 -- thanks for this. Looks good. 

Let me have a play. 

I think we should probably go with it as an option for folks. 

Nice.

channels_redis/alt.py
acu192: Efficient broadcast happens in a tree-structure. One connection to Redis *per consumer* is not a tree structure and will over-burden your Redis server. In-memory operations are always faster than network operations. You are essentially proposing that we replace in-memory operations with more network I/O. It's not a good idea. I'm closing this discussion. Please trust your voice has been heard. Next I want to hear from the maintainers of this repo.

channels_redis/alt.py
acu192: Again, one connection to Redis *per consumer* is not efficient. Imagine you had one million consumers connected to one thousand django server replicas. Now imagine the network I/O for that if you also had one million connections to the *same* Redis server. It would be WAY BETTER to have only one thousand connections to Redis (one per server replica) and each server Replica did some internal broadcasting in-memory.

channels_redis/alt.py
acu192: > Having local caches completely destroys this purpose.

No, local in-memory operations are ALWAYS BETTER than network I/O. Always.

The fastest channel layer you could ever have is actually the [InMemoryChannelLayer](https://github.com/django/channels/blob/588f7d3c7d0734727ffdd358d5b48145b9979258/channels/layers.py#L192). But it doesn't work if you need more than one Django server replica. Therefore, we use Redis to fill in the gaps ... communication between replicas. That's what this PR does.

Please know your voice has been heard and we have discussed this enough. Next let me discuss with the maintainers.

acu192:
@carltongibson Thanks for considering this PR! I'm excited to get it integrated and help maintain it over time.

If you want a simple project to play around with this, I copied this implementation into this repo:
https://github.com/acu192/channels-bug

You can easily change which channel layer is used [on this line of code](https://github.com/acu192/channels-bug/blob/6ec3012bb4a6d9eba09f559a3e1aa50393953b07/mysite/settings.py#L43). Currently it is set to use this new implementation, as you'll see. Maybe that gives you an easy way to play around with it.

acu192:
@carltongibson Also, after this is merged, I'll create issues in this repo to document the features that are still needed in this implementation to bring it in-line with the original implementation. I'd it's best if we address those additional features in future PRs, to take this incrementally.

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+        Includes '__group__' in the returned
+        string so that these names are distinguished
+        from those returned by `new_channel()`.
+        Technically collisions are possible, but it
+        takes what I believe is intentional abuse in
+        order to have colliding names.
+        """
+        return f"{self.prefix}__group__{group}"
+
+    extensions = ["groups"]
```
yedpodtrzitko, line 55: OT: what is this for? When I search for its usage I cant find any. I can see it's defined in all other `ChannelLayer` classes too, but I cant figure out its purpose. Asgiref defined something like `extensions` in its docs but it seems just like a name conincidence (?)

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+            # be named `delete_channel()`. If that were the case, we would do the
+            # following cleanup from that new `delete_channel()` method, but, since
+            # that's not how Django Channels works (yet), we do the cleanup below:
+            if channel in self.channels:
+                del self.channels[channel]
+                try:
+                    shard = self._get_shard(channel)
+                    await shard.unsubscribe(channel)
+                except Exception:
+                    traceback.print_exc(file=sys.stderr)
```
yedpodtrzitko, line 109: wouldnt be better to use `logging.exception()` here?

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+            # would serve as the antithesis of `new_channel()`; this new method might
+            # be named `delete_channel()`. If that were the case, we would do the
+            # following cleanup from that new `delete_channel()` method, but, since
+            # that's not how Django Channels works (yet), we do the cleanup below:
+            if channel in self.channels:
+                del self.channels[channel]
+                try:
+                    shard = self._get_shard(channel)
+                    await shard.unsubscribe(channel)
+                except Exception:
```
yedpodtrzitko, line 108: shouldnt this be `BaseException`, so a potential `CancelledError` from `unsubscribe()` is also caught to not suppress the original `CancelledError`?

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+        if channel not in self.channels:
+            raise RuntimeError(
+                "You can only call group_add() on channels that exist in-process.\n"
+                "Consumers are encouraged to use the common pattern:\n"
+                f"   self.channel_layer.group_add({repr(group)}, self.channel_name)"
+            )
+        group_channel = self._get_group_channel_name(group)
+        if group_channel not in self.groups:
+            self.groups[group_channel] = set()
+        group_channels = self.groups[group_channel]
```
yedpodtrzitko, line 132: nitpick: `group_channels` and `group_channel` are named too similarly to make a mistake easily.

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+        Return the connection to this shard that is used for *publishing* messages.
+
+        If the connection is dead, automatically reconnect.
+        """
+        if self._lock is None:
+            self._lock = asyncio.Lock()
+        async with self._lock:
+            if self._pub_conn is not None and self._pub_conn.closed:
+                self._pub_conn = None
+            if self._pub_conn is None:
```
yedpodtrzitko, line 212:
```suggestion
            while self._pub_conn is None:
```
^ this could replace the `while True:` + `break` below (?)

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+                while True:
+                    try:
+                        self._pub_conn = await aioredis.create_redis(self.host)
+                        break
+                    except Exception:
```
yedpodtrzitko, line 217:
```suggestion
                    except BaseException:
```
to catch a potential `CancelledError`?

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+                        print(
```
yedpodtrzitko, line 218:
```suggestion
                        logging.exception(
```

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+            elif name in self.channel_layer.groups:
+                for channel_name in self.channel_layer.groups[name]:
+                    if channel_name in self.channel_layer.channels:
+                        self.channel_layer.channels[channel_name].put_nowait(message)
+
+    async def _do_keepalive(self):
+        """
+        This task's simple job is just to call `self._get_sub_conn()` periodically.
+
+        Why? Well, calling `self._get_sub_conn()` has the nice side-affect that if
```
yedpodtrzitko, line 288:
```suggestion
        Why? Well, calling `self._get_sub_conn()` has the nice side-effect that if
```

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+        Adds the channel name to a group.
+        """
+        if channel not in self.channels:
+            raise RuntimeError(
+                "You can only call group_add() on channels that exist in-process.\n"
+                "Consumers are encouraged to use the common pattern:\n"
+                f"   self.channel_layer.group_add({repr(group)}, self.channel_name)"
+            )
+        group_channel = self._get_group_channel_name(group)
+        if group_channel not in self.groups:
```
yedpodtrzitko, line 130: `self.groups = defaultdict(set)` instead of a dict in constructor to get rid of this `if`?

yedpodtrzitko:
nicely done. I have just a few nitpicks about the code, feel free to disregard them if you dont consider them important.

Would it be possible to have a basic test added too, please?

channels_redis/alt.py
acu192: Honestly I only included it because I saw it in other `ChannelLayer`s (like you also noticed). E.g. [here](https://github.com/django/channels/blob/588f7d3c7d0734727ffdd358d5b48145b9979258/channels/layers.py#L217). Maybe someone, somewhere uses it...

channels_redis/alt.py
acu192: Yes, I agree. Although we don't have access to an instantiated `Logger` object here. We could create our own logger (as in `logging.getLogger(__name__)`) but I don't like that either. It would be best to hook into an existing logger (like Uvicorn and/or Daphne each instantiate a logger, it would be great to have access to _it_ here). Thoughts?

channels_redis/alt.py
acu192: Yes, I agree. I'll make this change.

channels_redis/alt.py
acu192: We've debated `defaultdict` in my team, and came to the conclusion that most of the time it leads to more confusion code (even thought it is *less* code; less `if`s). I'll make a longer comment about this on one of your other comments. At the end of the day, it's just a matter of preference, to be fair.

channels_redis/alt.py
```diff
@@ -0,0 +1,311 @@
...
+            self.groups[group_channel] = set()
+        group_channels = self.groups[group_channel]
+        if channel not in group_channels:
```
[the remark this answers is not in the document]
acu192: > this `if` is not really necessary, is it?

You are correct, this `if` is *not* necessary. But I like it. I think it communicates better the idea that "this thing might already be here" to the reader. It's a matter of preference... a stylistic thing. If the maintainers of the project prefer a different style, I'll defer to them though.

This is similar to choosing to not use a `defaultdict`. I like being explicit with the dictionary/set operations. Using `defaultdict` reduces the lines of code but it creates more mental effort when reading the code. Again, just my opinion.

channels_redis/alt.py
acu192: Yes, I agree. Your suggestion is better. I'll make this change.

channels_redis/alt.py
acu192: I agree. I'll make the change.

channels_redis/alt.py
acu192: Same about the `Logger` object. Don't think we should create a new logger. Wish we could access the one used by Uvicorn/Daphne ... maybe there's a way?

channels_redis/alt.py
acu192: Good catch!

channels_redis/alt.py
yedpodtrzitko: > We could create our own logger (as in logging.getLogger(__name__))

Sounds good to me. It's done the same way in here in `channels_redis/core.py`, similarly in Channels themselves, so it would be nicely consistent. 

Plus in case someone is using Sentry etc., this exception wouldn't be collected as it is now.

acu192: Okay I'm convinced. I'll do it like `channels_redis/core.py` does.

channels_redis/alt.py
yedpodtrzitko: Could you please elaborate why do you think we shouldnt create a new logger? afaik it's the standard way how to do logging. It's done this way in Django, Channels and Channels-Redis too. So even if I would have a different personal preference about a certain coding practice, following the existing practice have a bigger weight.

Also not everyone is using file based logs, so tools which rely on `logging.exception` or `logging` in general would miss this one. Sentry is one of the most popular, but I met other ways how logs were collected and processed, and all of them always relied on `logging`.

acu192:
@yedpodtrzitko Your suggestions were fixed in the two commits above. Thank you!

channels_redis/alt.py
yedpodtrzitko: I saw it there too when I was searching for its purpose, but it doesnt seems to be used. I made it an empty list and everything was still working the same 🤔 

Maybe I'll make a MR to Channels to remove it from there to see what will be the reaction.

carltongibson (commented):
Hi @acu192 — thanks for this. It looks nice. 

Sorry for the delay. It's been hectic. 

Some initial comments: 

* I'm going to pull in #250. Can you look out for that and rebase? 
* `alt.py` is probably better as `pubsub.py` no? 
* For tests, do you think we can `parametrize` the `test_core` tests? (i.e. we just want to check the channel layer API works, so running base tests on both channel layers seems desirable.) 

Current plan is to go over the open PRs and then pull together a release. Thanks!

acu192:
@carltongibson The first two tasks are done:
1. This PR is rebased. It now includes the changes from the other PR (#250), which TBH, is *not* what I expected (maybe I did it wrong). If that's a problem I can close this PR and open a new one that is "fresh" with just my changes.
2. I agree. I renamed `alt.py` to pubsub.py`.

For the tests, I'll look into that next...

acu192:
Okay, looking at the tests there are a few internal (non-API) differences that make this hard to use the same exact test code, for example the use of `.client_prefix` and `.close_pools()`.

Another issue is that this pub/sub impl requires that you call `.new_channel()` before you call `.receive()`. This is because we "subscribe" to specific channels from inside `.new_channel()`. This works fine in production (given how Django Channels uses the API) but many of the tests in `test_core` do not do this.

Another issue is that Redis Pub/Sub does not have a unique keyspace per databases, thus the trick that `test_core` uses for multi-database testing will not work here. See [this documentation](https://redis.io/topics/pubsub), section "Database & Scoping" which says:
> Pub/Sub has no relation to the key space. It was made to not interfere with it on any level, including database numbers. ... Publishing on db 10, will be heard by a subscriber on db 1.

There may be other issues, but I stopped at this point. How about I write a new test suite (`tests/test_pubsub.py`) to test this new impl and ensure it implements the API correctly (now and in the future)?

carltongibson:
>  How about I write a new test suite (tests/test_pubsub.py) to test this new impl and ensure it implements the API correctly (now and in the future)?

OK, that sounds perfect. (The thought to parametrise was just about saving a bit of work, but yes, it’s not at the right level for that.)

Please, yes, perhaps create fresh PR, as there’s a lot of noise here now — I think you merged, rather than rebased at the end there — and a fresh one-or-two-commit PR will be easier to review. 

Thanks! (This is quite exciting 😄)

acu192:
Here is a link to the new (cleaned-up) PR for this feature: https://github.com/django/channels_redis/pull/251

PR #255, opened 2021-05-26
Add default to serialize method

sevdog:
This is a revised version of #237. Instead of adding an explicit default callbak to `msgpack.packb` a placeholder method is placed.

In this way it can be overriden according to use cases.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -832,11 +832,20 @@ class RedisChannelLayer(BaseChannelLayer):
 
     ### Serialization ###
 
+    def default_serialize(self, value):
+        """
+        Gets called for objects that can’t otherwise be serialized.
+        It should return a JSON encodable version of the object or raise TypeError.
+
+        It should be overriden when an alternative default serialization is required.
+        """
+        raise TypeError
+
     def serialize(self, message):
         """
         Serializes message to a byte string.
         """
-        value = msgpack.packb(message, use_bin_type=True)
+        value = msgpack.packb(message, use_bin_type=True, default=self.default_serialize)
         if self.crypter:
             value = self.crypter.encrypt(value)
 
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1,5 +1,6 @@
 import asyncio
 import random
+import uuid
 
 import async_timeout
 import pytest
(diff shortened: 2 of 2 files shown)


sevdog:
After 1132fccbaf924f705321dae69bb53ff0d13ac81d I have added support for pubsub serialize "default".

acu192:
I think it may be better to encourage people to override `serialize()` and `deserialize()` instead. Especially in the pubsub impl, those methods are [trivial](https://github.com/django/channels_redis/blob/fe27f1a83a87bf61a9245188c8825ddfa30119fa/channels_redis/pubsub.py#L57-L67) (each is only one line), so overriding them is easy to reason about.

sevdog:
@acu192 this was initially implemented long before the pubsub implementatio was developed and merged, when the only serialize method was that of `RedisChannelLayer` which is not trivial. I have updated this also with the pubsub to provide a more consistent API.

The data-type problem is intrinsic with the serialization process. Msgpack is a JSON-like serialization and many data types are not JSON-serializable without transformation (ie: date/time). This may bring problems when you try to realize a complex project and may provide arbitrary data to the channel layer, thus having something which can safe-convert this kind of data is required and the easier way to hook this is via the [`default`](https://msgpack-python.readthedocs.io/en/latest/api.html#msgpack.Packer) parameter (which is the equivalent of that in [`json.dump`](https://docs.python.org/3/library/json.html?highlight=default#json.dump)).

If you look to an other python project which handle multi-process communication (which is what channels does thanks to the underlying layer) like celery they provide a "default" handler for JSON serialization, thus avoiding unwanted errors in the serialization process. It is up the the developer to know how the project is configured and to choose the approach which fits better to his requirements.

I agree that also the serialize method could be overridden, if it is trivial, but when it is not like that it just bring more unwanted complexity to a project which would like to use this package without having to look inside it but just reading docs.

If you mind type the only serialization which can preserve them in python is pickle. But I am not going neither to suggest nor to use it when handling data which comes from an untrusted source (ie: the user).

This PR may be further improved, because I may think that there should be a common base class for every reids-channel-layer with a basic set of APIs, which may define a "default" hook for JSON-like serialization and other common methods.

Let me know what do you think, so I can improve this.

PR #257, opened 2021-06-25
Add sentinel support to pubsub layer

qeternity:
As the title says, this migrates the raw connections over to use aioredis pools which we acquire connections from, and in doing so gives us sentinel compatibility.


acu192:
This all looks fine to me.

carltongibson:
Ok, super.

I'll take where we are as a good point for the release, and start pulling g together release notes and such. (We can pull in further points as follow-ups I think, which means we don't need to rush any design decisions.)

Thanks both!

PR #258, opened 2021-06-25
Better pubsub event loop handling

qeternity:
This PR changes the shard logic to operate with event loop context in situations where multiple event loops are running.

### DIFF
--- a/channels_redis/pubsub.py
+++ b/channels_redis/pubsub.py
@@ -20,6 +20,7 @@ class RedisPubSubChannelLayer:
             isinstance(hosts, list) and len(hosts) > 0
         ), "`hosts` must be a list with at least one Redis server"
 
+        self.hosts = hosts
         self.prefix = prefix
 
         # Each consumer gets its own *specific* channel, created with the `new_channel()` method.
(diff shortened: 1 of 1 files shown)


carltongibson (commented):
Hey @qeternity — this has an intermittent error with a full channel. First failed on 3.7, then 3.9.

qeternity:
> Hey @qeternity — this has an intermittent error with a full channel. First failed on 3.7, then 3.9.

These are #229 in the core layer

acu192:
I _think_ this leads to duplicate messages if a group contains consumers in different event loops.

1. Consumer A (in event loop A) subscribes to "my group".
2. Consumer B (in event loop B) subscribes to "my group".
3. (now we have two `_do_receiving()` tasks (one per event loop) running and listening for "my group".
4. Messages arrives to "my group"
5. Both `_do_receiving()` tasks get it, both push it into the consumers' `Queue`s.
6. (now same message is pushed twice into each Queue)

Correct me if I'm wrong.

I think we need to keep a `self.channels` for each event loop, and a `self.groups` for each event loop ... to solve it.

Or create a new class (`class MultiLoopRedisPubSubChannelLayer`) that creates separate `RedisPubSubChannelLayer` objects, one per event loop. Like you are doing now, but just going one step higher so that you end up with one `RedisPubSubChannelLayer` for each event loop.

Thoughts?

qeternity:
@acu192 yes, absolutely right...that's what I get for trying to bang out late night POCs while I had some time. I'll circle back to this next week with fresh eyes and more time.

qeternity:
Actually - this becomes really prickly once you start allowing for loop closures...

PR #259, opened 2021-06-26
Add optional redis event handlers

qeternity:
What do we think about an approach similar to this? Keeps the reconnection/resubscribe logic but optionally exposes the events to consumers who want to implement handlers.

```
CHANNEL_LAYERS = {
    "default": {
        "BACKEND": "channels_redis.pubsub.RedisPubSubChannelLayer",
        "CONFIG": {
            "hosts": [...],
            "on_disconnect": "redis.disconnect",
        },
    },
}
```

And then in your channels consumer, you can implement the handler:

```
    async def redis_disconnect(self, *args):
        await self.close(WS_CODE_SERVICE_RESTART)
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,9 +51,8 @@ The server(s) to connect to, as either URIs, ``(host, port)`` tuples, or dicts c
 Defaults to ``['localhost', 6379]``. Pass multiple hosts to enable sharding,
 but note that changing the host list will lose some sharded data.
 
-Sentinel connections require dicts conforming to `create_sentinel
-<https://aioredis.readthedocs.io/en/v1.3.0/sentinel.html#aioredis.sentinel.
-create_sentinel>` with an additional `master_name` key specifying the Sentinel
+Sentinel connections require dicts conforming to `create_sentinel <https://aioredis.readthedocs.io/en/v1.3.0/sentinel.html#aioredis.sentinel.create_sentinel>`_
+with an additional `master_name` key specifying the Sentinel
 master set. Plain Redis and Sentinel connections can be mixed and matched if
 sharding.
 
@@ -165,6 +164,32 @@ If you're using Django, you may also wish to set this to your site's
         },
     }
 
+``on_disconnect`` / ``on_reconnect``
+~~~~~~~~~~~~
+
+The PubSub layer, which maintains long-running connections to Redis, can drop messages in the event of a network partition.
+To handle such situations the PubSub layer accepts optional arguments which will notify consumers of Redis disconnect/reconnect events.
+A common use-case is for consumers to ensure that they perform a full state re-sync to ensure that no messages have been missed.
+
+.. code-block:: python
+
+    CHANNEL_LAYERS = {
+        "default": {
+            "BACKEND": "channels_redis.pubsub.RedisPubSubChannelLayer",
+            "CONFIG": {
+                "hosts": [...],
+                "on_disconnect": "redis.disconnect",
+            },
+        },
+    }
+
+
+And then in your channels consumer, you can implement the handler:
+
+.. code-block:: python
+
+    async def redis_disconnect(self, *args):
+        # Handle disconnect
 
 Dependencies
 ------------
--- a/channels_redis/pubsub.py
+++ b/channels_redis/pubsub.py
@@ -13,7 +13,9 @@ class RedisPubSubChannelLayer:
     Channel Layer that uses Redis's pub/sub functionality.
     """
 
-    def __init__(self, hosts=None, prefix="asgi", **kwargs):
+    def __init__(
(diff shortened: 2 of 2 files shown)


acu192:
Is it easy enough to rebase this so the diff doesn't include the sentinal changes?

qeternity:
Sorry - not sure how I managed that. Anyway reverted now.

acu192:
This looks good to me. 

Since there aren't automated tests, have you tested it "by hand" (like, by restarting Redis on localhost) to make sure it does what we expect? I imagine it does... but just want to get confirmation of a by-hand test before merging.

acu192:
Also let's document this in the README.

qeternity:
> Since there aren't automated tests, have you tested it "by hand" (like, by restarting Redis on localhost) to make sure it does what we expect? I imagine it does... but just want to get confirmation of a by-hand test before merging.

Yes, it works in my very basic test env. I've never simulated a service failure in ci/cd before so not even sure how to do this. The easiest way is probably to create two redis instances, and then manually have the sentinel failover with a `SENTINEL failover master_name` command.

> Also let's document this in the README.

Done. I'll revisit the third PR next week as that's going to be a blocker to us getting this into our test env.

acu192:
Thanks @qeternity. I'm inclined to merge this, because it will detect _most_ (but not _all_) cases where message(s) were dropped. Indeed, _most_ is better than _none_. So for that reason I'd vote we merge it.

My only hesitation is misleading people to believe it's a full-proof solution (i.e. that it detects _all_ cases where a message is dropped). Maybe it's just a matter of communicating on the README that it is _still possible_ for messages to be dropped silently _in the case your app losses connection to the Redis server_. Maybe @carltongibson can weigh in?

This fact stems from the fact that a message pushed down into the network stack memory is not guaranteed to be delivered. Here's a concrete situation (IMO the _most likely_ situation where messages might be silently dropped).

1. You have a server replica A. It has a connection to Redis.
2. You have another server replica B. It has a _different_ connection to Redis.
3. Replica B's connection dies for whatever reason (maybe replica B's process is killed by the OS, or maybe a cat eats through the Ethernet cable which connects B to Redis, or whatever).
4. The messages pushed into the network stack on B never reach Redis.
5. Replica A maintains its connection to Redis. Nothing seems weird from the perspective of A.
6. Replica B comes back online and reconnects to Redis. It has _no idea_ which of its messages never reached the Redis server.
7. ... In the end, zero or more messages were dropped, but A never realized anything bad happened, and B cannot know which (if any) were dropped.

Two solutions:
1. Maybe... somehow... B comes back online and tries to notify A that it should just refresh everything. This is hard. Maybe B _never_ comes back online. Or maybe B _can't_ notify A because... well the network is the issue and the network is also the way that B would notify A.
2. Sequence numbers. This is the only correct solution IMO for when you **absolutely need to know for certain** if messages were dropped or not.

carltongibson:
Thanks both, I see no problem adding this. 

I'm currently just pondering exactly what we say to folks, about the choice of channel layer, and the limitations of that. (I also want to make sure we point to the RabbitMQ implementation in the docs too)

My plan is to draft something up tomorrow and then ask for your inputs.

acu192:
Thanks @carltongibson. Cool, I'll merge it then. 

Thanks @qeternity for so quickly jumping in to improve this pubsub impl. It's great to have this type of collaboration. 🤓

qeternity:
@acu192 Ok so this is targeting issues on the redis side of things, not in the producers/consumers themselves. That is probably outside the scope of this library and a difficult problem in all realms. However I'm not sure your characterization of redis is correct. I'm new to redis pubsub so I may be wrong: but redis will ack your publish commands, so if you absolutely need to guarantee that it was acked by redis, you could (attempt to) implement an at-least-once deliverability check. It's not quite as fire-and-forget as simply moving some bytes into the network stack and going about your business.

This functionality is better described above where you have uvicorn pods A and B. Pod A suffers a network blip for whatever reason to the redis server, and now want to instruct all of its connected consumers to perform a complete re-sync.

The situation you're describing indeed requires handling in the application itself, as it's going to be extremely context specific.

acu192:
Yes, I hear you. I like this PR's feature. It's a simple way to detect and recover from a Redis reboot or whatever.

My concern is only about how we communicate/document the guarantees of this channel layer (or _any_ channel layer that uses Redis). I.e. sending a message to Redis (even with an _ack_) does _not_ guarantee it gets delivered to the subscribers on the other side. That's in contrast to (for example) RabbitMQ's [durable queue](https://www.rabbitmq.com/queues.html#durability) where getting an _ack_ from the server actually means it is persisted by the server (and continues to be persisted until ack'd by the consumer(s)). Anyway, I know we agree. I just want some version of this rant documented somewhere, so that's why I'm bring it up so much.

PR #260, opened 2021-07-01
Version 3.3.0 Release. 

carltongibson:
Hi @acu192 @qeternity 

reparing the 3.3. release. 

Thus far just:

* Updated Readme to mention PubSub layer.

Did either of have thoughts on additional content/warnings/etc? (ASGI is explicitly _At most once delivery_. I'm not sure how much we need to labour that. If folks need guaranteed delivery going to need to program for it... Thoughts? 🤔)


qeternity:
Howdy @carltongibson - yes I think the pubsub layer just needs to have some clarification around the current difference in behavior from the core layer: The two that come to mind are:

1. anyone sensitive to redis interruptions and subsequent missed messages needs to implement the consumer notifications
2. the current event loop handling is not as forgiving as the core layer, so certain integrations (i.e. Celery) will not be drop-in replacements until we can improve that

And a massive thanks to @acu192 for contributing and putting some momentum behind this project!

carltongibson:
Change notes are for that last item. 😀

carltongibson:
@qeternity I added change notes in 9b103a5625ae391aa11da6679b57089de4f433a8. I think they more of less cover it. 

I think the README could probably do with a restructure — separating the options for the separate channel layers, but I think better to release and work on that.

carltongibson:
OK, bumped version. 

Unless you, @qeternity and @acu192, have input, I'll tag and release later on. 

Thanks both for the great additions here.

acu192:
Thanks @carltongibson for prepping this release. Exciting stuff!

> I think the README could probably do with a restructure — separating the options for the separate channel layers, but I think better to release and work on that.

Agreed, we can tackle that for next release, perhaps.

carltongibson:
Opened #261 to track that. 

Doesn't have to wait for a release. It can just be merged in any time.

PR #262, opened 2021-07-02
PubSub Layer Proxy

qeternity:
Ok folks - what do we think about this? Due to PubSub, we need to isolate the channels queues across connections, and the simplest way to do so is a layer per loop. On the other hand, this is still much more efficient than the Core layer because we can handle many channel receivers with a single Sub connection.

For performance reasons, I'd much rather address this at the `get_channel_layer()` level but not sure that's feasible.

I am deploying this into testing env tonight and will see how it fares.

### DIFF
--- a/channels_redis/pubsub.py
+++ b/channels_redis/pubsub.py
@@ -1,5 +1,8 @@
 import asyncio
+import functools
 import logging
+import sys
+import types
 import uuid
 
 import aioredis
@@ -8,7 +11,68 @@ import msgpack
 logger = logging.getLogger(__name__)
 
 
+if sys.version_info >= (3, 7):
+    get_running_loop = asyncio.get_running_loop
+else:
+    get_running_loop = asyncio.get_event_loop
+
+
+def _wrap_close(proxy, loop):
+    original_impl = loop.close
+
+    def _wrapper(self, *args, **kwargs):
+        if loop in proxy._layers:
+            layer = proxy._layers[loop]
+            del proxy._layers[loop]
+            loop.run_until_complete(layer.flush())
+
+        self.close = original_impl
+        return self.close(*args, **kwargs)
+
+    loop.close = types.MethodType(_wrapper, loop)
+
+
 class RedisPubSubChannelLayer:
+    def __init__(self, *args, **kwargs) -> None:
+        self._args = args
+        self._kwargs = kwargs
+        self._layers = {}
+
+    def __getattr__(self, name):
+        if name in (
+            "new_channel",
+            "send",
+            "receive",
+            "group_add",
+            "group_discard",
+            "group_send",
+            "flush",
+        ):
+            return functools.partial(self._proxy, name)
+        else:
+            return getattr(self._get_layer(), name)
+
+    def _get_layer(self):
+        loop = get_running_loop()
+
+        try:
+            layer = self._layers[loop]
+        except KeyError:
+            layer = RedisPubSubLoopLayer(*self._args, **self._kwargs)
+            self._layers[loop] = layer
+            _wrap_close(self, loop)
+
+        return layer
+
+    def _proxy(self, name, *args, **kwargs):
+        async def coro():
+            layer = self._get_layer()
+            return await getattr(layer, name)(*args, **kwargs)
+
+        return coro()
+
+
+class RedisPubSubLoopLayer:
     """
     Channel Layer that uses Redis's pub/sub functionality.
     """
--- a/tests/test_pubsub.py
+++ b/tests/test_pubsub.py
@@ -5,6 +5,7 @@ import async_timeout
 import pytest
 from async_generator import async_generator, yield_
 
+from asgiref.sync import async_to_sync
 from channels_redis.pubsub import RedisPubSubChannelLayer
 
 TEST_HOSTS = [("localhost", 6379)]
@@ -34,6 +35,18 @@ async def test_send_receive(channel_layer):
 
 
 @pytest.mark.asyncio
+def test_send_receive_sync(channel_layer, event_loop):
+    _await = event_loop.run_until_complete
+    channel = _await(channel_layer.new_channel())
+    async_to_sync(channel_layer.send, force_new_loop=True)(
+        channel, {"type": "test.message", "text": "Ahoy-hoy!"}
+    )
+    message = _await(channel_layer.receive(channel))
+    assert message["type"] == "test.message"
+    assert message["text"] == "Ahoy-hoy!"
+
+
+@pytest.mark.asyncio
 async def test_multi_send_receive(channel_layer):
     """
     Tests overlapping sends and receives, and ordering.
@@ -45,6 +58,19 @@ async def test_multi_send_receive(channel_layer):
     assert (await channel_layer.receive(channel))["type"] == "message.1"
     assert (await channel_layer.receive(channel))["type"] == "message.2"
     assert (await channel_layer.receive(channel))["type"] == "message.3"
+
+
+@pytest.mark.asyncio
+def test_multi_send_receive_sync(channel_layer, event_loop):
+    _await = event_loop.run_until_complete
+    channel = _await(channel_layer.new_channel())
+    send = async_to_sync(channel_layer.send)
+    send(channel, {"type": "message.1"})
+    send(channel, {"type": "message.2"})
+    send(channel, {"type": "message.3"})
+    assert _await(channel_layer.receive(channel))["type"] == "message.1"
+    assert _await(channel_layer.receive(channel))["type"] == "message.2"
+    assert _await(channel_layer.receive(channel))["type"] == "message.3"
 
 
 @pytest.mark.asyncio
@@ -101,3 +127,12 @@ async def test_random_reset__channel_name(channel_layer):
     channel_name_2 = await channel_layer.new_channel()
 
     assert channel_name_1 != channel_name_2
+
+
+def test_multi_event_loop_garbage_collection(channel_layer):
+    """
+    Test loop closure layer flushing and garbage collection
+    """
+    assert len(channel_layer._layers.values()) == 0
+    async_to_sync(test_send_receive)(channel_layer)
+    assert len(channel_layer._layers.values()) == 0
--- a/tests/test_pubsub_sentinel.py
+++ b/tests/test_pubsub_sentinel.py
@@ -5,6 +5,7 @@ import async_timeout
 import pytest
 from async_generator import async_generator, yield_
 
+from asgiref.sync import async_to_sync
 from channels_redis.pubsub import RedisPubSubChannelLayer
 
 SENTINEL_MASTER = "sentinel"
@@ -35,6 +36,18 @@ async def test_send_receive(channel_layer):
 
 
 @pytest.mark.asyncio
+def test_send_receive_sync(channel_layer, event_loop):
+    _await = event_loop.run_until_complete
+    channel = _await(channel_layer.new_channel())
+    async_to_sync(channel_layer.send, force_new_loop=True)(
+        channel, {"type": "test.message", "text": "Ahoy-hoy!"}
+    )
+    message = _await(channel_layer.receive(channel))
+    assert message["type"] == "test.message"
+    assert message["text"] == "Ahoy-hoy!"
+
+
+@pytest.mark.asyncio
 async def test_multi_send_receive(channel_layer):
     """
     Tests overlapping sends and receives, and ordering.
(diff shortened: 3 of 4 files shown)


channels_redis/pubsub.py
```diff
@@ -8,7 +9,51 @@
...
+
 class RedisPubSubChannelLayer:
+    def __init__(self, *args, **kwargs) -> None:
+        self._args = args
+        self._kwargs = kwargs
+        self._layers = {}
+
+    def __getattr__(self, name):
+        try:
+            loop = asyncio.get_event_loop()
```
carltongibson, line 32: get_event_loop() is deprecated in Python 3.10.

carltongibson (commented):
Hey @qeternity — Thanks for this. 

Can I ask you to outline (perhaps afresh) what the purpose of this is? What's the issue, and how does this resolve it? 

Looking at the new layer, I'm thinking it satisfies the core use-case perfectly well. In contrast spinning up an event loop per-connection seems... well... _unnecessarily complex_, shall we say. As such, I think I'm missing your concern. 

Thanks.

qeternity:
@carltongibson this doesn't create an event loop per connection, but rather it scopes the channel layer to an event loop. Our use-case is that we offload a lot of event processing to our celery stack, and we then emit channels messages from those tasks. In contrast with the rabbitmq layer, redis connections are quite cheap and so this ends up being a good tradeoff. However, due to celery's prefork model, we have to deal with multiple event loops. The proxy shim provides the right accessor to a given pool of pubsub queues/shards dependent on the event loop.

This isn't a hill I want to die on. I'm happy to maintain an internal fork as we've done before. However anybody hoping to use this layer with celery will be unable to do so (and must use the core layer).

I threw this together last night and wasn't able to deploy to our test env until this morning. I am trying to pin down a `Cannot run the event loop while another loop is running` in the loop close handler.

channels_redis/pubsub.py
qeternity: Thanks - resolved in latest.

acu192:
Hey sorry for the delay, I just noticed this PR today.

Without running/testing this, it *looks like* it will work fine.

I would suggest we keep the original `RedisPubSubChannelLayer` as-is (because I think it will satisfy >95% of the users' use-cases), then we can provide this PR's idea as a new class with a different name.

Name idea for the new class? Maybe `MultiAsyncioLoopRedisPubSubChannelLayer` ... kinda long, but at least it is descriptive of the purpose.

qeternity:
Thanks @acu192 - just to be clear, this was very much a proof-of-concept (it passed `channels_redis` unit tests here but we have yet to put it through its paces) to address some of the messier setups that the core layer will cooperate with.

acu192:
In light of https://github.com/django/channels_redis/issues/266#issuecomment-875801433 this PR is a new priority, so that things "work" no matter if people use Daphne or Uvicorn.

LiteWait:
@qeternity I tried this PR and the issue mentioned in #266 still happens when the async_to_sync is called. Not sure how this is happening given you only have a single event loop now per layer.  I'll try to get some time to dig deeper in the next few days. Oh and @acu192 this is happening running Uvicorn as well so I don't think the issue is related to just Daphne.  Also, as I have said **I just can't break your original PR**.

qeternity:
@LiteWait given your usage to `async_to_sync` I wouldn't expect this PR to work for you. This is an RFC - we haven't even tested this yet. It's simply to gather thoughts on an approach to building a proxy shim in this manner. It's also very unlikely that the original PR worked for you, but rather you just weren't seeing any exceptions.

qeternity:
@LiteWait Give the latest push a try, it should work even with your aggressive `async_to_sync` usage.

I force pushed a few times to rerun unit tests, but that completely unrelated `ChannelFull` error keeps cropping up in the core layer. Can @carltongibson or @acu192 please re-run?

carltongibson:
@qeternity I've re-run. Started to keep an eye on https://github.com/django/channels_redis/issues/229 — only a couple, but it's the same test so likely just bumping the capacity would help us there.

carltongibson (commented):
We're going to need to document this. Why/when/how. 

Do we have _some_ testing method beyond "Does it work in your production environment"? 
🤔

qeternity:
> We're going to need to document this. Why/when/how.

Definitely. This PR is just meant to get the discussion going. Lots of work to do before a _possible_ merge, and even more before marking the PubSub layer GA.

> Do we have _some_ testing method beyond "Does it work in your production environment"?
> 🤔

I will put together some tests that are invoking through some common multi-loop entry points (`async_to_sync` likely being the most common)

qeternity:
Ok so I've just (very hastily) run all of the existing tests through `async_to_sync` with the addition of a `_sync` variant for each test (not copied for sentinel tests yet). I've also added in a test asserting that we flush and clean up any layers upon event loop closure.

Would be great to see if @acu192 can do some benchmarking of the shim overhead as seems they have the most supporting infra around testing the pubsub behavior. (I'm also happy to have two layers so that people who don't invoke from any frankenstein sync contexts don't need to incur the overhead).

acu192:
I'm beginning to wonder if this is possible or if we need to take a step back and think about another way (not sure *what* way, yet). Consider `async_to_sync(group_add)(...)`... in the case it *does* creates a new event loop (which is the trouble case), it then creates a new Redis channel layer for that new loop (which is this PR's idea), then subscribes to the group, then closes the loop (after `group_add` returns) which destroys the Redis channel layer. At the end of the day we've "handled" the `group_add` call and cleaned up that new loop and its resources, but your consumer isn't actually subscribed to that group (since everything go cleaned-up). Right?

If I am right about the issue above, we need to do something else. Maybe the Redis channel layer has a single loop and if you call into it from a *different* loop it delegates the call into its own loop (via `loop.call_soon()` or even `loop.call_soon_threadsafe()` if the caller doesn't have a loop, which means if the caller forgets `async_to_sync` it could still work, which is a nice side-effect).

We know the channel layer will get created with a top-level, persistent event loop (see [here](https://github.com/django/channels/blob/433e9689095b223bb17b70b1db3a3cc4043c00bb/channels/consumer.py#L44)), so that can be the loop the everything else gets delegated to (in the case someone calls into the channel layer from a *different* loop or even from a thread with no loop).

So then your shim wouldn't create extra "sub-channel layers" (as it does now). Instead it would just check which loop your in (or if you have no loop) and call the correct asyncio funtion to delegate the call into the "correct" loop (i.e. the one loop that the channel layer expects). Just an idea... not sure if it's the right way yet. Thoughts?

PS: Thanks @LiteWait for testing the new impl and surfacing this bug. Thanks @qeternity for diving in on this PR so quickly. It's a tough problem for sure.

qeternity:
@acu192 Yes, all of this is correct. See in OP `I'd much rather address this at the get_channel_layer()` for this exact reason. We will get a first loop upon lazy channel creation, but I wouldn't call it permanent. If, for whatever reason, the original loop is closed, then the entire channel layer will die as all roads lead to Rome (even if unlikely). This is going to be a (relatively) fundamental limitation of pubsub, similar to the out of process `group_add` mentioned in another issue.

When I first looked at your layer, I had originally began building a single loop inside the layer in which all the shards would live, but I quickly realized this was fraught with issues (and I was just trying to quickly hack something together so that we could get the layer into testing). Basically, I couldn't figure a performant way to execute in another event loop (without resorting to some gross `ThreadPoolExecutor(max_workers=1)` type stuff). But I didn't really spend too long on it, as the approach in this PR yielded results pretty quickly.

I don't want to micro-optimize for our use, but in a typical stack, most of the subscribers will exist within consumers in long-lived processes/event loops and `group_add` invocations will almost always occur in the same process and event loop (or be subject to the current out of process limitations). However the publishers might be dispatched through a task queue, an `async_to_sync` inside of an API call, etc in which case we don't care about the publisher being long lived. Of course there will be exceptions to every rule, but I expect the above to be the overwhelming majority of people.

qeternity:
@LiteWait can you give this a try please

LiteWait:
@qeternity I am away on vacation will try it immediately when I get back. Thanks for getting on this.

LiteWait:
@qeternity I managed to try this.  It seems to be working as expected, however I am getting a warning (which makes no sense to me) from asgiref in the async_to_sync() call:

**UserWarning: async_to_sync was passed a non-async-marked callable**

The callable name is functools.partial(<bound method RedisPubSubChannelLayer._proxy of <erc_framework.pubsub.RedisPubSubChannelLayer object at 0x7f550bbdf240>>, 'send') 

send() is an async method so I assume functools.partial is causing the issue?

I assume this is harmless.

qeternity:
Yes, it's just a broken type check (which I'm pretty sure is fixed in recent versions). Ugly but harmless.

acu192:
@LiteWait According to [this answer](https://stackoverflow.com/a/52422903/4194814), Python 3.8+ won't show that warning.

__

@qeternity 

> I don't want to micro-optimize for our use, but in a typical stack, most of the subscribers will exist within consumers in long-lived processes/event loops and `group_add` invocations will almost always occur in the same process and event loop ...

What about the case when a synchronous consumer (i.e. subclass of `SyncConsumer` or `WebsocketConsumer`) use `group_add`?

As a concrete example from [the Channels Tutorial](https://channels.readthedocs.io/en/stable/tutorial/part_2.html) (scroll way down on that page), here is a code snippet from that tutorial which does what I'm talking about:

```python
class ChatConsumer(WebsocketConsumer):
    def connect(self):
        self.room_name = self.scope['url_route']['kwargs']['room_name']
        self.room_group_name = 'chat_%s' % self.room_name

        # Join room group
        async_to_sync(self.channel_layer.group_add)(
            self.room_group_name,
            self.channel_name
        )

        self.accept()

...
```

So, it does `async_to_sync(self.channel_layer.group_add)(...)`, which I'm worried about *in the case* that `async_to_sync` creates-then-destroys a new event loop. I feel like I need to dig deeper into `async_to_sync` to know *which cases* cause it to create-destroy a new event loop... but I'll just bringing it up here in the meantime. @qeternity Do you know off-hand if this will be an issue? I'm worried this case won't work because it will add the consumer to the group momentarily (while the temporary event loop exists), but the consumer is removed from the group as the call returns.

acu192:
From a little experimenting I did last week, it seems that running under Uvicorn, `async_to_sync` is smart enough to find-and-use the top-level event loop that Uvicorn creates.

But, again with my limited experiments, running under Daphne gives different results: `async_to_sync` creates-then-destroys an event loop in every call.

qeternity:
I'm traveling right now so unfortunately not very useful. But aren't all SyncConsumers ultimately wrapped in a `sync_to_async`? In that case I'm pretty sure `async_to_sync` uses the thread local loop.

But going back to an earlier comment: I think the silver bullet has to be the local event loop which does everything. Otherwise I don't think we'll be able to get away from some of these pubsub realities (I'm also not sure how big of an issue they are in practice...so far the other feedback has been consistent with my own, which this PR fixes).

FWIW we are running in staging, this exact commit. Seems to be running very well.

acu192:
> But aren't all SyncConsumers ultimately wrapped in a `sync_to_async`? In that case I'm pretty sure `async_to_sync` uses the thread local loop.

Yes... very good point, okay that makes sense; I'm feeling comfortable about it now. Thank you.

> FWIW we are running in staging, this exact commit. Seems to be running very well.

Happy to hear that!

acu192:
I'm pretty convinced this is a correct solution. You already have tests passing (✔️), so that's done.

I'll put this in staging on our site. Then if the three of us (me, @qeternity, @LiteWait) are all seeing good things after some amount time, let's merge it and do a point release.

LiteWait:
I can report, other than the asgiref warning I am seeing, this code has passed all my tests. I plan run it on my test sites early next week.

LiteWait:
@qeternity @acu192 Just an update, I've been running this the last 6 days, working perfectly.

qeternity:
@LiteWait glad to hear.

We deployed into staging last week, and did a small canary deploy early this week. Everything seems to working smoothly. We will review canary log data at the end of this week and likely deploy to prod next week.

qeternity:
Canary deploy looks good.

@acu192 We are seeing an issue (which is rather benign) around `group_discard` because the PubSub layer does some additional checking that the Core layer doesn't do (and the comments in the Core layer suggest that this is a conscious design choice). I think that we'll have to remove the assertions and just have some conditional checking instead.

```
    async def group_discard(self, group, channel):
        """
        Removes the channel from a group.
        """
        group_channel = self._get_group_channel_name(group)
        assert group_channel in self.groups
        group_channels = self.groups[group_channel]
        assert channel in group_channels
        group_channels.remove(channel)
        if len(group_channels) == 0:
            del self.groups[group_channel]
            shard = self._get_shard(group_channel)
            await shard.unsubscribe(group_channel
```

acu192:
On our end everything is looking good with this. No noticeable differences from the original `pubsub.py` version. So, since all three of us are happy and the tests pass, let's do it.

@qeternity I agree. Let's remove those asserts and replace with conditionals. Want to do a quick PR for that?

acu192:
Also, thank you @qeternity for this PR. Takes it to the next level for sure!

LiteWait:
@acu192 @qeternity update: working flawlessly, once we have a new release plan to put it into production ASAP. Appreciate all your efforts since I could only offer myself as a guinea pig here.

Off-Topic: I am working to migrate from Daphne to Uvicorn.  Are you running Uvicorn or Gunicorn with Uvicorn workers?  I have NGINX proxying to...
`uvicorn --uds blah.sock --forwarded-allow-ips * --no-use-colors --proxy-headers asgi:channel_layer`
Not sure how that translates to Gunicorn if you guys recommend it.

revoteon:
@carltongibson When is this planned to be released?

carltongibson:
@revoteon — I've been slightly occupied with the Django 4.0 feature freeze, but it's now back on my list to do a release of channels_redis.

carltongibson:
OK, there's an intermittent failure coming up from this PR. See https://github.com/django/channels_redis/pull/275#issuecomment-925653653 — if anyone has a cycle to dig into that it would be great.

PR #263, opened 2021-07-06
Fix RuntimeError when modifying in_use map

### LINKED ISSUE #243: RuntimeError: dictionary keys changed during iteration
Sometimes I can see the following error in the logs:

      ...
      File "/usr/local/lib/python3.8/site-packages/channels_redis/core.py", line 122, in close_loop
        for k, v in self.in_use.items():
    RuntimeError: dictionary keys changed during iteration

I don't use the latest redis_channel package, but I can see the same loop construct in current master as well:
 
https://github.com/django/channels_redis/blob/38ffdeb610ce175a01698ca267a6cccc171043c9/channels_redis/core.py#L125

        for k, v in self.in_use.items():
            if v is loop:
                self.in_use[k] = None

According to the python3 docs:

       Iterating views while adding or deleting entries in the dictionary 
       may raise a RuntimeError or fail to iterate over all entries. 

I'm happy to send a patch to set key as `None` out of for loop, although I'm a bit confused by the construct itself.

jalaziz:
Python throws a RuntimeError if we try to modify a dictionary while
iterating over a view of its items. We avoid the error by iterating over
a copy of the items in the dictionary.

Also, esnure that the in_use dictionary does not contain empty mappings
by deleting the entry when closing the loop.

fixes #243

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -103,8 +103,7 @@ class ConnectionPool:
         """
         Return a connection to the pool.
         """
-        loop = self.in_use[conn]
-        del self.in_use[conn]
+        loop = self.in_use.pop(conn, None)
         if loop is not None:
             conns, _ = self._ensure_loop(loop)
             conns.append(conn)
@@ -114,7 +113,7 @@ class ConnectionPool:
         Handle a connection that produced an error.
         """
         await self._close_conn(conn)
-        del self.in_use[conn]
+        self.in_use.pop(conn, None)
 
     def reset(self):
         """
@@ -143,10 +142,10 @@ class ConnectionPool:
                 await self._close_conn(conn)
             del self.conn_map[loop]
 
-        for k, v in self.in_use.items():
+        for k, v in list(self.in_use.items()):
             if v is loop:
                 await self._close_conn(k)
-                self.in_use[k] = None
+                self.in_use.pop(k, None)
 
     async def close(self):
         """


si14:
I haven't got to the bottom of it, but I do get this error with this PR (on 3.3.0 I get the error from #243):

```
self = <channels_redis.core.ConnectionPool object at 0x7fe5673d7550>, loop = <_UnixSelectorEventLoop running=False closed=False debug=True>

    async def close_loop(self, loop):
        """
        Close all connections owned by the pool on the given loop.
        """
        if loop in self.conn_map:
            for conn in self.conn_map[loop]:
                await self._close_conn(conn)
            del self.conn_map[loop]
    
        for k, v in list(self.in_use.items()):
            if v is loop:
                await self._close_conn(k)
>               del self.in_use[k]
E               KeyError: <Redis <ConnectionsPool [db:0, size:[1:1], free:0]>>

/home/djangousr/.local/lib/python3.8/site-packages/channels_redis/core.py:148: KeyError
```

The reason I get it might be somewhere else, but I thought it might be helpful.

jalaziz:
Interesting. Seems like maybe `self.in_use` is being modified indirectly by `self._close_conn`.

Will try to confirm and update with a better fix.

Ian-Foote:
I can confirm I see the same `KeyError` with this patch.

Would a patch that just wrapped `self.in_use.items()` in `list` (or `tuple`) be sufficient?

jalaziz:
@si14  pushed a fix that should solve the error.

However, I haven't been able to figure out why it would happen. Do you have to have reproduction steps so I could try and debug locally?

channels_redis/core.py
```diff
@@ -102,9 +102,9 @@ def push(self, conn):
         """
         Return a connection to the pool.
         """
-        loop = self.in_use[conn]
-        del self.in_use[conn]
+        loop = self.in_use.get(conn)
```
yedpodtrzitko, line 105:
```suggestion
        loop = self.in_use.pop(conn, None)
```

jalaziz: Indeed it does. Also removed another instance of `del` just to avoid a potential error there too.

carltongibson (commented):
Wondering if we can put together a test that reliably fails without the patch... 🤔

jalaziz:
> Wondering if we can put together a test that reliably fails without the patch... 🤔

I've been wondering it too. I see the issue periodically when running unit tests and I haven't been able to figure out exactly why it happens. Seems related to tear down, but it's not consistent.

jalaziz:
@carltongibson I finally figured out the root cause. Just need to see if I can write a test.

In my case, the root cause is effectively a race condition between `conn_error` and `close_loop`.

If there's a connection disconnect/error around the same time that the loop is closing, there's a chance that `conn_error` deletes the connection from `in_use` as `close_loop` is looping through `in_use`. This can happen because `close_loop` calls `await self._close_conn(k)`, relinquishing control back to the loop allowing `conn_error` to delete the key from `in_use` first. When control is given back to `close_loop`, the `for` loop finishes the iteration and then immediately errors out.

carltongibson:
@jalaziz super, good work. 🏅

jalaziz:
So far no luck reproducing it with a simple test.

I've been playing with different variations of the following using the `243-debug` branch on my fork:
```
@pytest.mark.asyncio
async def test_close_loop_keyerror(event_loop, channel_layer):
    async def _connection_error():
        async with channel_layer.connection(0) as connection:
            raise Exception()

    async def _close_loop():
        await channel_layer.pools[0].close_loop(event_loop)
    
    async with channel_layer.connection(0) as connection:
        pass # Do nothing

    conn_error_task = asyncio.create_task(_connection_error())
    close_loop_task = asyncio.create_task(_close_loop())
    await asyncio.wait([conn_error_task, close_loop_task])
```

What's odd is I'm seeing the same sequence of debug print statements as when I was able to reproduce it reliably through a test in our downstream codebase that was using communicators. Will keep playing around, but sharing the details here in case I'm missing something obvious.

Also, if it helps anyone, the issue started appearing reliably when we upgraded to Python 3.10.

In the test that reproduced it reliably, we had two `WebSocketCommunicator` instances and an `ApplicationCommunicator` instance. The test that was failing was calling `communicator.connect()` on one of the websocket communicators, but was calling `communicator.disconnect()` on all of them at the end (disconnecting a communicator that was never connected). Fixing the cleanup code to only call `disconnect` on the connected communicator fixed the test.

carltongibson:

PR #267, opened 2021-07-16
Fix RuntimeError: dictionary keys changed during iteration

Ian-Foote:
Simpler fix for #243 than #263.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -142,7 +142,7 @@ class ConnectionPool:
                 await self._close_conn(conn)
             del self.conn_map[loop]
 
-        for k, v in self.in_use.items():
+        for k, v in tuple(self.in_use.items()):
             if v is loop:
                 await self._close_conn(k)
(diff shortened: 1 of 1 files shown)


ngnpope:
👋🏻 @carltongibson 

I've been using the fix here by monkey patching `ConnectionPool.close_loop()`.

I notice that `channels_redis.core.ConnectionPool` is no longer present in `4.0.0b1`, so could this fix be released in a `3.4.1`? One reason to make such a release is that, while it makes sense to migrate to the new async support in `redis-py`, some may decide to hold off on upgrading as that async support is based on `aioredis>=2.0.0` and there are still some unaddressed performance concerns.

ngnpope:
Sure. Could you create a `3.x` branch at the `3.4.0` tag for me to target?

carltongibson:
I went with 3.4.x, since I'm **not** maintaining this long term. 😄

carltongibson:

PR #274, opened 2021-09-01
Fix #273 -- Use consistent hashing for PubSub

### LINKED ISSUE #273: PubSub: Sharding broken with multiple workers or worker restarts
The "classic" implementation selects shards like this:

```
    def consistent_hash(self, value):
        """
        Maps the value to a node value between 0 and 4095
        using CRC, then down to one of the ring nodes.
        """
        if isinstance(value, str):
            value = value.encode("utf8")
        bigval = binascii.crc32(value) & 0xFFF
        ring_divisor = 4096 / float(self.ring_size)
        return int(bigval / ring_divisor)
```

The PUBSUB implementation does this:

```
    def _get_shard(self, channel_or_group_name):
        """
        Return the shard that is used exclusively for this channel or group.
        """
        if len(self._shards) == 1:
            # Avoid the overhead of hashing and modulo when it is unnecessary.
            return self._shards[0]
        shard_index = abs(hash(channel_or_group_name)) % len(self._shards)
        return self._shards[shard_index]
```

I was surprised that there's a difference when I first looked at this, but now this started to really bite us in production after we introduced sharding: `hash()` is not guaranteed to return the same thing across multiple invocations of `python`, even on the same machine, unless `PYTHONHASHSEED` is set manually, which probably no-one ever does. See https://docs.python.org/3/using/cmdline.html#envvar-PYTHONHASHSEED

I think we should just use the same hash function across implementations. I'll send a PR :)

raphaelm:


carltongibson (commented):
@raphaelm Thanks for this!

Just wondering about tests... 🤔

raphaelm:
Any specific ideas? Maybe a test with a hardcoded shard index that needs to be matched?

carltongibson:
Yes... — it's trying to frame a test which is _telling_ on `_get_shard()`, or higher up... Not sure. 🤔

Can you look at the lint failure: 

> channels_redis/pubsub.py:114:18: F821 undefined name 'binascii'

acu192:
This looks fine except for the missing import (see linter error).

It's hard to figure out how to test this (from the test suite). The code short-circuits with only one shard (which I'd argue is something we should keep), so that code you are adding is not covered by any of the test suite.

We'd have to have two instances of Redis running in the test suite in order to test with sharding. We cannot have one Redis instance and use _keyspaces_ with pub/sub: See [this documentation](https://redis.io/topics/pubsub), section "Database & Scoping" which says:
> Pub/Sub has no relation to the key space. It was made to not interfere with it on any level, including database numbers. ... Publishing on db 10, will be heard by a subscriber on db 1.

So... how hard would it be to run two instances of Redis server in the test suite? At a glance I don't know.

carltongibson:
So worst case is we’re not loosing coverage, but we don’t have a regression test as it stands…

raphaelm:
> We'd have to have two instances of Redis running in the test suite in order to test with sharding. We cannot have one Redis instance and use keyspaces with pub/sub: See this documentation, section "Database & Scoping" which says:

But we *can* run two channel

acu192:
This is a big issue so maybe @raphaelm can fix the linter issue then hand-test this new code block and we can merge and release. As @carltongibson points out, we are not *losing* coverage at least.

raphaelm:
Sorry about the missing import, will fix tomorrow. Was late doing the PR ;) I'll also look into testing and manual tests again, but we're successfully running a patched version in production for a few days now :)

carltongibson:
I'm happy to push out a point release this week. 

@raphaelm If you can think of something for testing then super. Otherwise don't worry. (The code is copied from the older layer so...) 

Thanks for the input!

raphaelm:
I pushed one commit fixing the linter issue, and another one refactoring the hash function in a way that we only have it in the codebase once instead of twice. It also adds a test with a few static values to make sure the hash function doesn't change unexpectedly. Feel free to not merge this last commit if you don't like the introduction of a utils module, or let me know if I should make any changes.

carltongibson (approved):
Hey @raphaelm — I think that's just about right. Thanks! 

I'm flat out first half of the week. I'll get this out a release out after that. 

🎁

acu192:
@raphaelm - this looks great, thank you!

Seems there is one linter issue still (this time about style/formatting... run `black` on util.py and commit the changes to fix it)

After that we can merge! 🍾

raphaelm:
@acu192 Yeah, sorry 'bout that. A little annoying that actions don't run on PRs from first-time contributors, so I couldn't check the CI output right away. Hopefully it's done now ;)

acu192:
@raphaelm Thank you! 

Yes, [this is why](https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/) we can't have nice things.

Anyway, thank you for finding and fixing this! I'll merge next.

acu192:
The test failures now are just this old issue: https://github.com/django/channels_redis/issues/229

PR #250, opened 2021-05-12
Added Redis sentinel support.

carltongibson:
Supersedes #246 @qeternity — just edits, but I couldn't push to your PR.

PR #275, opened 2021-09-23
Prepare 3.3.1 release

carltongibson:


carltongibson:
Hey @qeternity — we've got an intermittant failure on one of the sentinel tests... 

```
tests/test_pubsub_sentinel.py::test_multi_send_receive_sync
165
  /home/runner/work/channels_redis/channels_redis/.tox/py39/lib/python3.9/site-packages/asgiref/sync.py:125: UserWarning: async_to_sync was passed a non-async-marked callable
166
    warnings.warn("async_to_sync was passed a non-async-marked callable")
167
```

This from the PY39 run, but the same error came up on PY36 and PY37 the first run.

The test was added in 49e419dce642b7aef12daa3fa8495f6b9beb1fbe for #262. Do you have any bandwidth to take a look? (Ta!)

revoteon:
The actual failure is in `test_receive_cancel`, that was added with sentinel support in #250. https://github.com/django/channels_redis/blob/f5eef165017a702a1e53fe19a673036c0e337c95/tests/test_sentinel.py#L438

I suspect this is related to the design of the test, which intermittently can go >= channel capacity and raise `ChannelFull` as per above. To replicate the issue more reliably, reduce the capacity to a low number (1 or 2), when initiating the layer in the test. 

Also, same intermittent error happens in `test_core.test_receive_cancel` as well, which is exactly the same design.

There is a comment  in `RedisChannelLayer.send` that indicates that going over capacity is possibility.
```python
# Check the length of the list before send
# This can allow the list to leak slightly over capacity, but that's fine.
```

carltongibson:
@revoteon — yes. See #229 — for the other failure. It just needs someone to sit down and adjust it, and I've yet to have the bandwidth... — but very happy to take input there if you have a cycle 🥳

(First item on my list is just to bump the capacity for the test, and see if that resolves it.)

carltongibson:
Rebased after #277.

PR #293, opened 2022-02-16
[Feature] Discard full channels

nick-merrill:
Hey there. I'd love to have this option included with Channels Redis, given the challenges some folks have with channels not always being discarded due to whatever other reasons. Thanks for your time!

## Problem

We have an issue where channels are no longer being consumed from, but they linger around anyway. This requires us to support far more channels than we need to.

Ideally we'd figure out how to discard these channels upon clients disconnecting, but even though we call `group_discard` appropriately, the channels don't seem to be discarded.

## Solution

Add an option that, when enabled, discards any at-capacity channels immediately upon sending a message that channel.

## Testing

- [x] Added an automated test
- [x] Tested with our own application code

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -143,6 +143,13 @@ to 10, and all ``websocket.send!`` channels to 20:
 If you want to enforce a matching order, use an ``OrderedDict`` as the
 argument; channels will then be matched in the order the dict provides them.
 
+``should_auto_discard_full_channels``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When this option is set to ``True`` and a message is sent to a channel that is at its
+maximum capacity (e.g. 100 messages are in a channel whose capacity is 100), the
+*entire channel* will be discarded from its group.
+
 ``symmetric_encryption_keys``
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -143,6 +143,13 @@ to 10, and all ``websocket.send!`` channels to 20:
 If you want to enforce a matching order, use an ``OrderedDict`` as the
 argument; channels will then be matched in the order the dict provides them.
 
+``should_auto_discard_full_channels``
```
nick-merrill, line 146: ℹ️ Note to self to update the changelog before merging.

nick-merrill:
Closing since there seems to be no interest in pulling this into the main branch, which I understand. Folks can always cherry-pick if they need to, but it could be worth considering adding some of the comments and code cleanup of the Luascript sometime.

carltongibson:
Hey @nick-merrill — still waiting for a moment to sit down and think clearly about I'm afraid. 🤹 

Thanks for the input!

nick-merrill:
No worries at all @carltongibson! Feel free to reopen if that works better for the organization of this project. I'll defer to you on that, as I would be quite lost organizing an open source project myself. Major kudos to you all for the great work!

PR #297, opened 2022-03-06
Fix iscoroutine check for pubsub proxied methods

jalaziz:
Adjust the proxying approach in `RedisPubSubChannelLayer` to ensure the
proxied methods are properly marked as coroutines according to
`inspect.iscoroutinefunction`. This can be issue during testing when
trying to mock channel layer methods.

The underlying issue is a [Python bug][1] where partial methods are not
fully unwrapped to determine if they are coroutines. The workaround
simply moves the proxy wrapper to a top-level async function, tricking
`inspect.iscoroutinefunction` to correctly detecting the proxied async
methods.

[1]: https://bugs.python.org/issue38364

### DIFF
--- a/channels_redis/pubsub.py
+++ b/channels_redis/pubsub.py
@@ -34,6 +34,13 @@ def _wrap_close(proxy, loop):
     loop.close = types.MethodType(_wrapper, loop)
 
 
+async def _async_proxy(obj, name, *args, **kwargs):
+    # Must be defined as a function and not a method due to
+    # https://bugs.python.org/issue38364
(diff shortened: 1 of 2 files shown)


jalaziz:
@qeternity Ran into this while trying to fix Channels workers with the pubsub layer. I unfurled the inner function because it didn't seem necessary (the proxy coroutine shouldn't be evaluated until called), but figured you might know better.

qeternity:
LGTM

carltongibson (approved):
Thanks @jalaziz 👍

PR #298, opened 2022-03-06
Fix workers support when using Redis PubSub layer

### LINKED ISSUE #270: Channels' worker fails to start with PubSub layer
Hello,

With the new pubsub layer, Channels worker fails to start due to the following error:

`RuntimeError: You should only call receive() on channels that you "own" and that were created with new_channel()`
from [here](https://github.com/django/channels/blob/main/channels/worker.py#L37)

I can patch the Worker's [handle function](https://github.com/django/channels/blob/main/channels/worker.py#L26) by replicating the logic of [.new_channel()](https://github.com/django/channels_redis/blob/main/channels_redis/pubsub.py#L146) for each channel but it seems like a hack.

Is it an expected behavior with the new pubsub logic? If not should it be fixed in channels or channels_redis ?

Thank you.

jalaziz:
The new Redis PubSub layer broke support for Channels workers. Add
support for workers by subscribing to non-owned channels instead of
throwing an exception.

fixes #270

### DIFF
--- a/channels_redis/pubsub.py
+++ b/channels_redis/pubsub.py
@@ -138,6 +138,11 @@ class RedisPubSubLoopLayer:
         """
         return f"{self.prefix}__group__{group}"
 
+    async def _subscribe_to_channel(self, channel):
+        self.channels[channel] = asyncio.Queue()
+        shard = self._get_shard(channel)
+        await shard.subscribe(channel)
+
     extensions = ["groups", "flush"]
 
     ################################################################################
@@ -157,9 +162,7 @@ class RedisPubSubLoopLayer:
         process as a specific channel.
         """
         channel = f"{self.prefix}{prefix}{uuid.uuid4().hex}"
-        self.channels[channel] = asyncio.Queue()
-        shard = self._get_shard(channel)
-        await shard.subscribe(channel)
+        await self._subscribe_to_channel(channel)
         return channel
 
     async def receive(self, channel):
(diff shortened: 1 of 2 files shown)


jalaziz:
While looking into this, I realized that the PubSub layer doesn't actually conform to the [Channels Layer Spec](https://channels.readthedocs.io/en/stable/channel_layer_spec.html).

In particular, the spec says:
> Channels are a first-in, first out queue with at-most-once delivery semantics. They can have multiple writers and multiple readers; only a single reader should get each written message. Implementations must never deliver a message more than once or to more than one reader, and must drop messages if this is necessary to achieve this restriction.

While the PubSub layer is "at-most-once", it doesn't guarantee that a message will only be delivered to a single reader. This has implications for workers as having multiple workers listening to the same channel will result in duplicate messages.

A potential solution is to introduce a Redis Streams layer which allows such guarantees.

acu192:
This looks good to me. (Except the PY36 failure, but I agree with @carltongibson that we should drop PY36 in the next release.)

jalaziz:
Put up #301 to drop PY36.

carltongibson:
@jalaziz Good work. If you want to rebase here I'll take another look tomorrow 🛏️ 🎁

jalaziz:
> @jalaziz Good work. If you want to rebase here I'll take another look tomorrow 🛏️ 🎁

Was already on it :)

carltongibson (approved):
Super, let's have it. Thanks (again) @jalaziz! 🏅

acu192:
Thank you @jalaziz! Great work 👍, and super valuable.

carltongibson:
I can roll a release this week I guess... — Anything else to clean up whilst we're here?

jalaziz:
> I can roll a release this week I guess... — Anything else to clean up whilst we're here?

There's the `RuntimeError` when the dictionary gets modified that we used to see, but that's not blocking anything. Not suer if there's anything that's easy to fix.

PR #306, opened 2022-04-11
Ping redis connection before using it

retupmoca:
If the redis server closes the connection, we may not know about that
closure until we attempt to use it and get a network or aioredis error.
If we ping-test the connection first, we can catch and reconnect without
causing any application code to receive errors.


qeternity:
This is no longer required as redis-py pools handle disconnects gracefully.

Issue #312, opened 2022-06-30
v4b1: RuntimeError: Event loop is closed


nielsuit227:
I'm not sure whether the root cause is in the 4.0.0 release or in Django-Celery's 5.3.0 release of two days ago, but since today, we've been getting a recurring error in our CI, and after updating the local environment, also locally. 

Before, all works well. With these two updates, I get the recurring error shown below. 

OS: Windows 10 & Ubuntu 20.04.4
Pip Freeze
```
alembic==1.8.0
amqp==5.1.1
APScheduler==3.9.1
argparse-addons==0.8.0
asgiref==3.5.2
async-timeout==4.0.2
atomicwrites==1.4.0
attrs==21.4.0
autobahn==22.5.1
Automat==20.2.0
autopage==0.5.1
azure-core==1.24.1
azure-storage-blob==12.12.0
billiard==4.0.0
bitstruct==8.15.1
boto3==1.24.20
botocore==1.27.20
cachetools==5.2.0
cantools==37.1.0
catboost==1.0.6
celery==5.3.0a1
certifi==2022.6.15
cffi==1.15.0
channels==3.0.5
channels-redis==4.0.0b1
charset-normalizer==2.1.0
click==8.1.3
click-didyoumean==0.3.0
click-plugins==1.1.1
click-repl==0.2.0
cliff==3.10.1
cloudpickle==2.1.0
cmaes==0.8.2
cmd2==2.4.1
colorama==0.4.5
colorlog==6.6.0
constantly==15.1.0
cryptography==37.0.2
cycler==0.11.0
daphne==3.0.2
defusedxml==0.7.1
Deprecated==1.2.13
diskcache==5.4.0
dj-rest-auth==2.2.4
Django==4.0.5
django-allauth==0.51.0
django-apscheduler==0.6.2
django-celery-beat==2.3.0
django-cors-headers==3.13.0
django-filter==22.1
django-redis==5.2.0
django-timezone-field==5.0
djangorestframework==3.13.1
faiss-cpu==1.7.2
fonttools==4.33.3
google-api-core==2.8.2
google-auth==2.9.0
google-cloud-core==2.3.1
google-cloud-storage==2.4.0
google-crc32c==1.3.0
google-resumable-media==2.3.3
googleapis-common-protos==1.56.3
graphviz==0.20
greenlet==1.1.2
gunicorn==20.1.0
hyperlink==21.0.0
idna==3.3
imbalanced-learn==0.8.0
incremental==21.3.0
[...]

PR #317, opened 2022-07-13
v4b1 - AsyncToSync closed event loop

bbrowning918:
- Re implement lazy pool generation for the running event loop
- Add test since to hopefully not regress the issue again

refs https://github.com/django/channels_redis/issues/312

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -122,8 +122,9 @@ class RedisChannelLayer(BaseChannelLayer):
         # Configure the host objects
         self.hosts = self.decode_hosts(hosts)
         self.ring_size = len(self.hosts)
-        # Cached redis connection pools
-        self.pools = self.create_pools()
+        # Cached redis connection pools and the event loop they are from
+        self.pools = {}
+        self.pools_loop = None
         # Normal channels choose a host index by cycling through the available hosts
         self._receive_index_generator = itertools.cycle(range(len(self.hosts)))
         self._send_index_generator = itertools.cycle(range(len(self.hosts)))
@@ -147,23 +148,19 @@ class RedisChannelLayer(BaseChannelLayer):
         # a message back into the main queue before its cleanup has completed
         self.receive_clean_locks = ChannelLock()
 
-    def create_pools(self):
-        pools = []
-        for host in self.hosts:
-            if "address" in host:
-                pools.append(aioredis.ConnectionPool.from_url(host["address"]))
-            elif "master_name" in host:
-                sentinels = host.pop("sentinels")
-                master_name = host.pop("master_name")
-                pools.append(
-                    aioredis.sentinel.SentinelConnectionPool(
-                        master_name, aioredis.sentinel.Sentinel(sentinels), **host
-                    )
-                )
-            else:
-                pools.append(aioredis.ConnectionPool(**host))
+    def create_pool(self, index):
+        host = self.hosts[index]
 
-        return pools
+        if "address" in host:
+            return aioredis.ConnectionPool.from_url(host["address"])
+        elif "master_name" in host:
+            sentinels = host.pop("sentinels")
+            master_name = host.pop("master_name")
+            return aioredis.sentinel.SentinelConnectionPool(
+                master_name, aioredis.sentinel.Sentinel(sentinels), **host
+            )
+        else:
+            return aioredis.ConnectionPool(**host)
 
     def decode_hosts(self, hosts):
         """
@@ -519,8 +516,8 @@ class RedisChannelLayer(BaseChannelLayer):
         # pools without flushing first.
         await self.wait_received()
 
-        for pool in self.pools:
-            await pool.disconnect()
+        for index in self.pools:
+            await self.pools[index].disconnect()
 
     async def wait_received(self):
         """
@@ -782,4 +779,16 @@ class RedisChannelLayer(BaseChannelLayer):
             raise ValueError(
                 "There are only %s hosts - you asked for %s!" % (self.ring_size, index)
             )
+
+        try:
+            loop = asyncio.get_running_loop()
+            if self.pools_loop != loop:
+                self.pools = {}
+                self.pools_loop = loop
+        except RuntimeError:
+            pass
+
+        if index not in self.pools:
+            self.pools[index] = self.create_pool(index)
+
         return aioredis.Redis(connection_pool=self.pools[index])
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -312,7 +312,7 @@ class RedisChannelLayer(BaseChannelLayer):
                 self.client_prefix + "!"
             ), "Wrong client prefix"
             # Enter receiving section
-            loop = asyncio.get_event_loop()
+            loop = asyncio.get_running_loop()
             self.receive_count += 1
             try:
                 if self.receive_count == 1:
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -404,6 +404,20 @@ async def test_group_send_capacity_multiple_channels(channel_layer, caplog):
         )
 
 
+def test_repeated_group_send_with_async_to_sync(channel_layer):
+    """
+    Makes sure repeated group_send calls wrapped in async_to_sync
+    process-local channel names.
+    """
+    channel_layer = RedisChannelLayer(hosts=TEST_HOSTS, capacity=3)
+
+    try:
+        async_to_sync(channel_layer.group_send)('channel_name_1', {"type": "test.message.1"})
+        async_to_sync(channel_layer.group_send)('channel_name_2', {"type": "test.message.2"})
+    except RuntimeError as exc:
+        pytest.fail(F"repeated async_to_sync wrapped group_send calls raised {exc}")
+
+
 @pytest.mark.asyncio
 async def test_receive_cancel(channel_layer):
     """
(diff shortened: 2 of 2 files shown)


bbrowning918:
I am not sure why the 3.7 tests hung, I will give them a try locally (did the work under 3.8)

carltongibson:
@bbrowning918 Cool. Thanks. 

From the log it looks like a hang. If you can’t resolve quickly, comment and I’ll have a look (rather than you spending an age on it!)

bbrowning918:
I could not get the tests to hang with either pytest with 3.7.13 or a tox run, but that's local. It seems it was the pub sub section, but no obvious clues as none of that changed.

bbrowning918:
I have gotten somewhere. The `test_groups_basic` in either `test_pubsub.py` and `test_pubsub_sentinel.py` tests can hang intermittently.

The part that hangs is the last section where we force a timeout for a message we correctly should not receive. It can be recreated by a test such as this, ran over and over until one sticks and never finishes.
```
@pytest.mark.asyncio
async def test_receive_timeout(channel_layer):
    channel_name = await channel_layer.new_channel(prefix="test-timeout-chan")
    with pytest.raises(asyncio.TimeoutError):
        async with async_timeout.timeout(1):
            await channel_layer.receive(channel_name)
```

Deeper inside `receive`, on the cleanup section triggered by `async_timeout` throwing `asyncio.CancelledError`, we attempt to unsubscribe the channel on the shard. 
```
async def unsubscribe(self, channel):
    if channel in self._subscribed_to:
        self._subscribed_to.remove(channel)
        print("looking for sub conn")
        conn = await self._get_sub_conn()
        print("found sub conn")
        await conn.unsubscribe(channel)
        print("unsubscribed")
```
`await self._get_sub_conn()` seems to await something that never arrives. That method is beyond my depth/knowledge.

CI environments could be more susceptible assuming they are more resource constrained than a dedicated machine locally, as it feels like a timing thing.

carltongibson:
OK, good detective work @bbrowning918 🕵️ 

@acu192 Can I ask you to have a glance if you get a cycle? (I shall have a dig too this week)

acu192:
@bbrowning918 Do you mind looking one call deeper to see which of the `await`s inside `_get_sub_conn()` is hanging?

I just spent some time looking (again) at the diff where we [migrated from aioredis to redis-py](https://github.com/django/channels_redis/commit/65a05285d512e32607829b45cba5b4093e6003a6#). I'm assuming that's where this creeped in, as we didn't see it before that change, and given _most_ of our [recent GitHub Actions have failed due to this](https://github.com/django/channels_redis/actions).

Beyond that... I'm struggling to keep my head above water so I can't look at this deeper right now. 😟

bbrowning918:
Okay, bear with the ignorance here but with more brute force printing I found the following.

```
    async def _get_sub_conn(self):
        if self._keepalive_task is None:
            self._keepalive_task = asyncio.ensure_future(self._do_keepalive())
        if self._lock is None:
            self._lock = asyncio.Lock()
        print(self._lock)
        async with self._lock:
            if self._sub_conn is not None and self._sub_conn.connection is None:
                await self._put_redis_conn(self._sub_conn)
                self._sub_conn = None
                self._notify_consumers(self.channel_layer.on_disconnect)
            if self._sub_conn is None:
                if self._receive_task is not None:
                    print(self._receive_task)
                    self._receive_task.cancel()
                    try:
                        print("waiting for receive_task")
                        await self._receive_task
                    except asyncio.CancelledError:
                        print("receive_task cancelled")
                        # This is the normal case, that `asyncio.CancelledError` is throw. All good.
                        pass
```
Which on a hanging run produces:

> <asyncio.locks.Lock object at 0x7f88fd85a7f0 [unlocked]>
<asyncio.locks.Lock object at 0x7f88fd85a7f0 [unlocked]>
<Task pending name='Task-4' coro=<RedisSingleShardConnection._do_receiving() running at channels_redis/channels_redis/pubsub.py:409> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f88fd895490>()]>>
waiting for receive_task
receive_task got cancelled
<asyncio.locks.Lock object at 0x7f88fd85a7f0 [unlocked]>
<Task pending name='Task-5' coro=<RedisSingleShardConnection._do_receiving() running at channels_redis/channels_redis/pubsub.py:391> wait_for=<Future cancelled>>
waiting for receive_task
<asyncio.locks.Lock object at 0x7f88fd85a7f0 [locked]>


The `_do_keepalive` is triggering sequential runs on it's periodic timer and we seem to hit a deadlock. We end up waiting at `self._recieve_task` for a second/different task but it's cancellation never seems to occur.

A successful run ends with the `CancellationError` thrown and all the tidying up done. We don't get that last keepalive attempt that sticks on the lock. 

> <asyncio.locks.Lock object at 0x7f49b6ab6550 [unlocked]>
<asyncio.locks.Lock object at 0x7f49b6ab6550 [unlocked]>
<Task pending name='Task-4' coro=<RedisSingleShardConnection._do_receiving() running at channels_redis/channels_redis/pubsub.py:408> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f49b6af5820>()]>>
waiting for receive_task
receive_task got cancelled
<asyncio.locks.Lock object at 0x7f49b6ab6550 [unlocked]>
<Task pending name='Task-5' coro=<RedisSingleShardConnection._do_receiving() running at channels_redis/channels_redis/pubsub.py:408> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f49b6af5c70>()]>>
waiting for receive_task
receive_task got cancelled

acu192:
I'm only coming up with more questions, like (from the code-path you are discovering) [this section](https://github.com/django/channels_redis/blob/d946ce2875b2a2efe2c4cbfb14107fd7d7a4a551/channels_redis/pubsub.py#L348-L351) is being hit... meaning that the Redis connection is lost _very consistently_ during this test (much more consistently than I'd expect, as we're seeing something like 4/5 of our recent github actions fail here). If that section were _not_ being hit, we'd never get into the deeper parts of the code that are blocking us. So... if I had time I'd actually start investigate _that question_.

But, onto the things you've discovered. You've discovered that `await self._receive_task` is the real blocker. If it were to return, the lock would _unlock_ and the `_do_keepalive()` would get to do its thing, and all would be well.

So (assuming we're both correct up until now), there are only two `awaits` inside `_do_receiving()`, [here](https://github.com/django/channels_redis/blob/d946ce2875b2a2efe2c4cbfb14107fd7d7a4a551/channels_redis/pubsub.py#L386-L388) and [here](https://github.com/django/channels_redis/blob/d946ce2875b2a2efe2c4cbfb14107fd7d7a4a551/channels_redis/pubsub.py#L404). Those are the only two place that can be blocking us. Umm, the latter is weird. Why are we sleeping here!? That should be removed. It slipped into the [migrated from aioredis to redis-py](https://github.com/django/channels_redis/commit/65a05285d512e32607829b45cba5b4093e6003a6#) somehow, although it's not causing _this particular_ issue. The former is all that's left, which (if I'm right), means that redis-py is actually what's blocking us.

All told, we should definitely move this to a new issue as it doesn't seem to relate to your PR here. In additions, while I'm dying to work on this (I really like this library and want to see it improve), I'm unable to commit significant time right now. I suspect (of course, unconfirmed) that we've done something wrong at a core level of our migration to redis-py -- not something that can be blamed by any _single_ line of code, but rather something about _how we're using redis-py_ that is breaking us. It's probably a deep issue.

acu192:
Edit: I immediately reevaluated my post, and made some strike-throughs on it after I realized that all the recent GH action failures are on _this_ PR. So, maybe the situation is _not_ so dire. Sorry for my error which led me to that rash conclusion.

acu192:
Edit edit: I decided to re-run the test suite in another PR, to see if this same deadlock would happen _elsewhere_. Indeed, it seems to have reproduced elsewhere, see: https://github.com/django/channels_redis/runs/7418417216  (I canceled the test after it had hung for over an hour)

So, I'm back to thinking this issue is unrelated to _this PR_, that it instead has something to do with our usage of redis-py. I'd love a second opinion on that though, as I've gone back and forth now.

bbrowning918:
I swapped locally back to main, reran test_receive_timeout a couple times and the hang reoccurred. I would agree this is a redis-py migration (intermittent) issue.

If this PR has to be set aside until the underlying hang/failure is solved, assuming this has not made it worse inadvertently, for some future merge in then that's just the way things go sometimes. A consistently passing test suite seems like the higher priority and harder problem for the redis-py bits to be completed. The underlying issue this PR sought to fix should not be difficult to do downstream of any other changes considering there is a simple test case to recreate/check.

If there is no objection, I will do my best to sum up what was learned so far and how to recreate the hang as a new issue for visibility sake with the hope more eyes on it will produce new insight or solutions.

acu192:
> If there is no objection, I will do my best to sum up what was learned so far and how to recreate the hang as a new issue for visibility sake with the hope more eyes on it will produce new insight or solutions.

Yes, please go ahead. Thank you for the hunting you've done so far. It's been very valuable.

carltongibson:
Thanks both for the work here! 🎁 

 I shall pick this up too and have a look.

bbrowning918:
The cleanup PRs either got some of the ghosts out, or there was finally a lucky test run.

carltongibson (approved):
OK, this seems a win, even if it's not **all** the issues here. Let's merge this, and simplify the state of play at least. 

Thanks @bbrowning918 !

Issue #328, opened 2022-09-19
Connection.disconnect: Task was destroyed but it is pending!


bellini666:
This is similar to https://github.com/django/channels_redis/issues/327 but happens when using `RedisPubSubChannelLayer`:

```
Task was destroyed but it is pending!
task: <Task pending name='Task-133' coro=<Connection.disconnect() running at /home/bellini/dev/2u/backend/.venv/lib/python3.10/site-packages/redis/asyncio/connection.py:837> wait_for=<Future finished result=None>
```

Like the other issue, it seems that it happens because the connection gets disconnected from the `__del__` method, but in this case the loop is already closed and cannot run it.

The fix here is trivial and I'm already opening a PR for it in a few minutes.

Issue #327, opened 2022-09-19
Connection.disconnect got Future attached to a different loop when sending a message in a django ASGI view


bellini666:
Just upgraded my project to 4.0.0b1 (with also updating channels/daphne/etc) and started to get this error after doing a `layer.group_send` inside a django ASGI view (i.e. not a channels consumer):

```
Task exception was never retrieved
future: <Task finished name='Task-153' coro=<Connection.disconnect() done, defined at /home/bellini/dev/2u/backend/.venv/lib/python3.10/site-packages/redis/asyncio/connection.py:822> exception=RuntimeError("Task <Task pending name='Task-153' coro=<Connection.disconnect() running at /home/bellini/dev/2u/backend/.venv/lib/python3.10/site-packages/redis/asyncio/connection.py:837>> got Future <Future pending> attached to a different loop")>
Traceback (most recent call last):
  File "/home/bellini/dev/2u/backend/.venv/lib/python3.10/site-packages/redis/asyncio/connection.py", line 837, in disconnect
    await self._writer.wait_closed()  # type: ignore[union-attr]
  File "/usr/lib/python3.10/asyncio/streams.py", line 344, in wait_closed
    await self._protocol._get_close_waiter(self)
RuntimeError: Task <Task pending name='Task-153' coro=<Connection.disconnect() running at /home/bellini/dev/2u/backend/.venv/lib/python3.10/site-packages/redis/asyncio/connection.py:837>> got Future <Future pending> attached to a different loop
```

It seems that this is scheduled from the `Connection.__del__` method.

obs. This does not break the view itself, but I see that message everytime. e.g. I have a post_save signal that will do a `group_send` to inform some listeners that the object has changes. When saving the object inside django admin or even in another view the `group_send` is executed correctly, the message is received by the listeners correctly and the view returns the response correctly, but I always see that traceback in my terminal

bbrowning918:
Does 4.0.0b2 also behave the same way? It has a few changes/fixes/improvements from b1 that look familiar to the above.

bellini666:
Hi @bbrowning918 ,

O sorry, that issue is in b2 actually (I got confused with `channels` itself, which is `b1`)

carltongibson:
Thanks for the report @bellini666. 

It's a cleanup warning... there's task not finished when the event loop shuts down. (You're running this under WSGI?)

A minimal reproduction would be handy.

carltongibson:
The trick will be to give the connection a chance to close before finishing.

[...]

PR #329, opened 2022-09-19
fix: Make sure the Connection.disconnect gets called before the loop is closed

bellini666:
Fix https://github.com/django/channels_redis/issues/328

### DIFF
--- a/channels_redis/pubsub.py
+++ b/channels_redis/pubsub.py
@@ -302,7 +302,10 @@ class RedisSingleShardConnection:
                     pass
                 self._receive_task = None
             if self._redis is not None:
-                await self._redis.close()
+                # The pool was created just for this client, so make sure it is closed,
+                # otherwise it will schedule the connection to be closed inside the
+                # __del__ method, which doesn't have a loop running anymore.
+                await self._redis.close(close_connection_pool=True)
                 self._redis = None
                 self._pubsub = None
             self._subscribed_to = set()
--- a/tests/test_pubsub.py
+++ b/tests/test_pubsub.py
@@ -32,6 +32,21 @@ async def other_channel_layer():
     await channel_layer.flush()
 
 
+def test_layer_close():
+    """
+    If the channel layer does not close properly there will be a "Task was destroyed but it is pending!" warning at
+    process exit.
+    """
+
+    async def do_something_with_layer():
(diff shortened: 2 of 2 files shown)


carltongibson:
Ah,a PR good... (didn't see this before commenting on the ticket just now 🙂)

How might we test this? 🤔

bellini666:
Hi @carltongibson ,

I tested it in a live project. Without this patch everytime I would instantiate the layer object and call `group_send` inside a django view (not a channel consumer) I would get the error I reported in the issue. That because the pool gets deleted inside the wrapped `loop.close`, which schedules `.disconnect()` to be called by the loop, but since it is being closed there it doesn't get a chance to run.

Wi that patch the error don't happen because the connection gets closed in that `redis.close()` call (which is the correct behaviour, since we don't have more than one connection in that pool. It was created specifically for that conn)

If you are looking for a unit test, I can try to write one. Don't know how to properly mock this to ensure that the coroutine was never run, but I can give it a try.

Note that this fixes the PubSub layer. The core one has a similar issue (https://github.com/django/channels_redis/issues/327), which I did not find a good solution for it. The only difference there is that the `disconnect` does get called, but it is somehow run by a different loop (probably because the [loop that created the pool](https://github.com/django/channels_redis/blob/main/channels_redis/core.py#L718) is not the same that called the [close_pools](https://github.com/django/channels_redis/blob/main/channels_redis/core.py#L492) function, but I don't know exactly why)

bellini666:
obs. I'm running the beta versions of channels and channels_redis, and while this is not merged I monkey-patched it to force the same behaviour:

```python
_original_ensure_redis = RedisSingleShardConnection._ensure_redis


def _ensure_redis(self):
    # FIXME: This can be removed once this issue is fixed and released:
    # https://github.com/django/channels_redis/issues/328
    need_patch_redis = self._redis is None
    retval = _original_ensure_redis(self)

    if need_patch_redis:
        self._redis.close = functools.partial(
            self._redis.close,
            close_connection_pool=True,
        )

    return retval


RedisSingleShardConnection._ensure_redis = _ensure_redis
```

It is running in production in a project that relies a lot in the PubSub behaviour for 3 days now and everything is running fine! :)

carltongibson:
Thanks for the follow up @bellini666   

As I say sounds correct. Let me have a play. 👍

PR #333, opened 2022-10-21
Update CHANGELOG.txt

vanschelven:
As is evidenced in 65a05285d512e32607829b45cba5b4093e6003a6 the (configuration) API is not entirely unchanged. In particular, using a tuple in the `address` section of the `hosts` `CONFIG` will now lead to a failure.


carltongibson (approved):
Yep, thanks.

Issue #331, opened 2022-09-30
Cannot connect to redis, using self-signed TLS and sentinels


ajgon:
I'm not sure if it's a bug or me being stupid (as I'm not a python developer), so sorry in advance :)

I'm trying to setup channels-redis (4.0.0b2, the one from main branch) with redis sentinel via SSL (self-signed) - SSL is both on sentinels and underlying redises. I have successfully managed to set up sentinels connection, and fetch master node - however I'm unable to join underlying redis. I'm getting CERTIFICATE_VERIFY_FAILED no matter what I try. This is what I'm trying to do:

```python
_ssl_context = ssl.create_default_context()
_ssl_context.check_hostname = False
_ssl_context.verify_mode = ssl.CERT_NONE

CHANNEL_LAYERS = {
  "default": {
    "BACKEND": "channels_redis.core.RedisChannelLayer",
    "CONFIG": {
      "hosts": [
        {
          "sentinels": [('sentinel.host', 26379)],
          "master_name": "mymaster",
          "sentinel_kwargs": {
            "password": "sentinel password",
            "ssl": True,
            "ssl_cert_reqs": "none",
          },
          "ssl": _ssl_context,
          "db": 0,
          "password": "redis password"
        }
      ],
    }
  }
}
```

and here is the exception:

```
  File "/usr/local/lib/python3.9/site-packages/redis/asyncio/connection.py", line 709, in connect
    await self._connect()
  File "/usr/local/lib/python3.9/site-packages/redis/asyncio/connection.py", line 744, in _connect
    reader, writer = await asyncio.open_connection(
  File "/usr/local/lib/python3.9/asyncio/streams.py", line 52, in open_connection
    transport, _ = await loop.create_connection(
  File "/usr/local/lib/python3.9/asyncio/base_events.py", line 1090, in create_connection
    transport, protocol = await self._create_connection_transport(
  File "/usr/local/lib/python3.9/asyncio/base_events.py", line 1120, in _create_connection_transport
    await waiter
  File "/usr/local/lib/python3.9/asyncio/sslproto.py", line 534, in data_received
    ssldata, appdata = self._sslpipe.feed_ssldata(data)
  File "/usr/local/lib/python3.9/asyncio/sslproto.py", line 188, in feed_ssldata
    self._sslobj.do_handshake()
  File "/usr/local/lib/python3.9/ssl.py", line 945, in do_handshake
    self._sslobj.do_handshake()
[...]

PR #337, opened 2022-11-02
Allow passing extra connection kwargs with host address

marcinowski:
This should fix some of the issues with #331 by allowing to pass extra connection arguments with the host configuration.
The specific case that this helps me with is passing the ssl kwargs into the connection as advised by [Heroku redis docs](https://devcenter.heroku.com/articles/connecting-heroku-redis#connecting-in-python)

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -130,7 +130,8 @@ class RedisChannelLayer(BaseChannelLayer):
         host = self.hosts[index]
 
         if "address" in host:
-            return aioredis.ConnectionPool.from_url(host["address"])
+            address = host.pop("address")
+            return aioredis.ConnectionPool.from_url(address, **host)
         elif "master_name" in host:
             sentinels = host.pop("sentinels")
             master_name = host.pop("master_name")
--- a/README.rst
+++ b/README.rst
@@ -65,13 +65,27 @@ Possible options for ``CONFIG`` are listed below.
 ``hosts``
 ~~~~~~~~~
 
-The server(s) to connect to, as either URIs, ``(host, port)`` tuples, or dicts conforming to `redis Connection <https://redis-py.readthedocs.io/en/v4.3.3/connections.html#redis.connection.Connection>`_.
+The server(s) to connect to.
+Can be either:
+
+- URIs with extra connection kwargs like
+
+.. code-block:: python
+
+    {
+        "address": "redis://localhost:6379",
+        "ssl_cert_reqs": None,
+    }
+
+
+- or ``(host, port)`` tuples
+- or dicts conforming to `redis Connection <https://redis-py.readthedocs.io/en/v4.3.3/connections.html#redis.connection.Connection>`_.
 Defaults to ``redis://localhost:6379``. Pass multiple hosts to enable sharding,
 but note that changing the host list will lose some sharded data.
 
 Sentinel connections require dicts conforming to:
 
-.. code-block::
+.. code-block:: python
 
     {
         "sentinels": [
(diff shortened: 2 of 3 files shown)


nasir733:
@carltongibson  what do you think about this pull request?

carltongibson:
@nasir733 I haven't had a chance to review it yes, otherwise I'd have commented. (@mentioning me doesn't help. I feel we've discussed this before... 😅) 

If you want to give it a run, verify it works (or not) for you and comment along those lines, then that would be a useful thing to do. (If it does work you're then able to use this branch whilst waiting for a release.)

bbrowning918:
Again with test hangs, I helped fight these not too long ago. I think with #342 also fixing pubsub's connection the two can be combined into one.

jackkinsella:
The approach in my PR seemed to just replace this issue with another: `TypeError: Connection.__init__() got an unexpected keyword argument 'ssl'

```
File "/app/.heroku/python/lib/python3.11/site-packages/channels/routing.py", line 116, in __call__
app[web.1]: return await application(
File "/app/.heroku/python/lib/python3.11/site-packages/channels/consumer.py", line 94, in app
return await consumer(scope, receive, send)
File "/app/.heroku/python/lib/python3.11/site-packages/channels/consumer.py", line 58, in __call__
await await_many_dispatch(
File "/app/.heroku/python/lib/python3.11/site-packages/channels/utils.py", line 57, in await_many_dispatch
await task
File "/app/.heroku/src/channels-redis/channels_redis/core.py", line 368, in receive
message_channel, message = await self.receive_single(
File "/app/.heroku/src/channels-redis/channels_redis/core.py", line 423, in receive_single
content = await self._brpop_with_clean(
File "/app/.heroku/src/channels-redis/channels_redis/core.py", line 259, in _brpop_with_clean
await connection.eval(cleanup_script, 0, channel, backup_queue)
File "/app/.heroku/python/lib/python3.11/site-packages/redis/asyncio/client.py", line 484, in execute_command
conn = self.connection or await pool.get_connection(command_name, **options)
File "/app/.heroku/python/lib/python3.11/site-packages/redis/asyncio/connection.py", line 1520, in get_connection
connection = self.make_connection()
File "/app/.heroku/python/lib/python3.11/site-packages/redis/asyncio/connection.py", line 1560, in make_connection
return self.connection_class(**self.connection_kwargs)
File "/app/.heroku/python/lib/python3.11/site-packages/redis/asyncio/connection.py", line 1088, in __init__
super().__init__(**kwargs)
TypeError: Connection.__init__() got an unexpected keyword argument 'ssl'
```

Perhaps a parallel with https://github.com/rq/rq/pull/1383

channels_redis/core.py
```diff
@@ -130,7 +130,8 @@ def create_pool(self, index):
         host = self.hosts[index]
 
         if "address" in host:
-            return aioredis.ConnectionPool.from_url(host["address"])
+            address = host.pop("address")
```
axieum, line 133: For this to work, we shouldn't mutate the original dict. We need to `deepcopy(host)`, and mutate that here.

ngnpope: Also see #341 which addresses the same mutation issue for the branch below.

It might make sense to just `copy.deepcopy()` on line 130 to avoid the risk of mutation in the downstream `redis` library too.

sevdog: This should also be addressed in the pubsub layer (note that pubsub already performs a _copy_ of host dict in `RedisSingleShardConnection` init):

https://github.com/django/channels_redis/blob/a7094c58a15cbf6e621b2129253060fc80cfdfad/channels_redis/pubsub.py#L351

PR #341, opened 2022-11-16
fix: django conf should not be modified

techblack:
This place is passed by reference, which will cause the configuration to change, and other applications will read the wrong configuration

before this pr:
`
from django.conf import settings
getattr(settings, "CHANNEL_LAYERS", {})
`
result

`{'default': {'BACKEND': 'channels_redis.core.RedisChannelLayer', 'CONFIG': {'hosts': [{'sentinels': [('a.b.c.d', 6379)], 'master_name': 'redis-master', 'db': 0, 'password': 'password', 'sentinel_kwargs': {'password': 'password'}}]}}}`

after channel connection result

`{'default': {'BACKEND': 'channels_redis.core.RedisChannelLayer', 'CONFIG': {'hosts': [{'db': 0, 'password': 'password'}]}}}`

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -11,7 +11,6 @@ import uuid
 
 import msgpack
 from redis import asyncio as aioredis
-from copy import deepcopy
 from channels.exceptions import ChannelFull
 from channels.layers import BaseChannelLayer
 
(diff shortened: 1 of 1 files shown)


krisatverbidio:
Also affected by this bug. Would love to see this get merged in soon.

carltongibson:
I'm planning releases for the new year period. This is on the agenda for that.

channels_redis/core.py
```diff
@@ -132,13 +132,14 @@ def create_pool(self, index):
         if "address" in host:
             return aioredis.ConnectionPool.from_url(host["address"])
         elif "master_name" in host:
-            sentinels = host.pop("sentinels")
-            master_name = host.pop("master_name")
-            sentinel_kwargs = host.pop("sentinel_kwargs", None)
+            host_kwargs = deepcopy(host)
+            sentinels = host_kwargs.pop("sentinels")
+            master_name = host_kwargs.pop("master_name")
+            sentinel_kwargs = host_kwargs.pop("sentinel_kwargs", None)
```
sevdog, line 138: There is no need to have an other variable (`host_kwargs`) since from this point there will be no more use of the old `host` variable.

To guard the host main dict from being changed it should be enough to use `.copy` method (as it is done in the pubsub layer), thus the deepcopy is not required.

https://github.com/django/channels_redis/blob/a7094c58a15cbf6e621b2129253060fc80cfdfad/channels_redis/pubsub.py#L266

Also a comment line may be useful here.

```suggestion
            # copy host dict to prevent original dict from being modified
            host= host.copy()
            sentinels = host.pop("sentinels")
            master_name = host.pop("master_name")
            sentinel_kwargs = host.pop("sentinel_kwargs", None)
```

sevdog (changes requested):
This should work like pubsub layer, if a deepcopy is realy needed it should also be addressed also in that layer.

sevdog (approved):

carltongibson:
Resolved in #352. Thanks.

PR #344, opened 2022-11-30
Makes group_discard a noop if channel not in group

### LINKED ISSUE #343: Mismatched `group_discard` behavior on RedisPubSub and core Redis layers
Versions:
- `channels` 4.0.0
- `channels-redis` 4.0.0

With `RedisChannelLayer`, `group_discard` explicitly does nothing if the channel is not in the group:
```
    async def group_discard(self, group, channel):
        """
        Removes the channel from the named group if it is in the group;
        does nothing otherwise (does not error)
        """
        assert self.valid_group_name(group), "Group name not valid"
        assert self.valid_channel_name(channel), "Channel name not valid"
        key = self._group_key(group)
        connection = self.connection(self.consistent_hash(group))
        await connection.zrem(key, channel)
```

In contrast, `RedisPubSubChannelLayer` will assert and fail if that's not the case:
```
    async def group_discard(self, group, channel):
        """
        Removes the channel from a group.
        """
        group_channel = self._get_group_channel_name(group)
        assert group_channel in self.groups
        group_channels = self.groups[group_channel]
        assert channel in group_channels
        group_channels.remove(channel)
        if len(group_channels) == 0:
            del self.groups[group_channel]
            shard = self._get_shard(group_channel)
            await shard.unsubscribe(group_channel)
```

For improved end-user expectations, they should both do the same thing. In my opinion, the original behavior of do nothing is more predictable and user-friendly, as any user of this library would have to otherwise functionally maintain `self.groups` on their own independently in order to verify that sending a `group_discard` method will not yield a crash.

If this change seems reasonable, I'm happy to put up a PR to this effect.

nbhargava:
Updates `RedisPubSubChannelLayer` to return early if `group_discard` is called with a channel that does not belong to the group.

Closes #343.

### DIFF
--- a/channels_redis/pubsub.py
+++ b/channels_redis/pubsub.py
@@ -223,12 +223,14 @@ class RedisPubSubLoopLayer:
 
     async def group_discard(self, group, channel):
         """
-        Removes the channel from a group.
+        Removes the channel from a group if it is in the group;
+        does nothing otherwise (does not error)
         """
         group_channel = self._get_group_channel_name(group)
-        assert group_channel in self.groups
-        group_channels = self.groups[group_channel]
-        assert channel in group_channels
+        group_channels = self.groups.get(group_channel, set())
+        if channel not in group_channels:
+            return
+
         group_channels.remove(channel)
         if len(group_channels) == 0:
             del self.groups[group_channel]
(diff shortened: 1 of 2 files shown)


acu192:
Thanks for finding this issue and fixing it.

nbhargava:
Would it be possible to get a patch version release with this change? We would love to stop surrounding every `group_discard` with a try-except if possible (though we can of course wait if it simplifies things for the maintainers).

carltongibson:
I'm planning updates for the new year period. This will be part of that. If you're in a rush you can use [pip's VCS Support](https://pip.pypa.io/en/stable/topics/vcs-support/#vcs-support) to install from GitHub.

Issue #348, opened 2023-02-28
redis-py >=4.4.0 breaking changes


bbrowning918:
This issue is to help with visibility and discussion:

redis-py >=4.4.0 introduces two breaking issues for the test suite and real usage on

1. `test_receive_cancel` in both core and sentinel will fail with "TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'" on the response to [ZREMRANGEBYSCORE](https://github.com/django/channels_redis/blob/main/channels_redis/core.py#L218). In usage it has been noted in redis-py on https://github.com/redis/redis-py/issues/2593 affecting normal looking `send` calls.
2. `test_message_expiry__group_send__one_channel_expires_message` can hang indefinitely in the test suite

These affect both unrelated changes/PRs and 4.0.0 in actual usage. The change log for redis-py 4.4.0 does not seem to have a smoking gun and pinning it to 4.3.5 has the test suite passing better.

carltongibson:
Thanks for writing this up @bbrowning918! 🏅

jajce:
I believe I tracked down and fixed the first issue you  described. I have a PR ready in redis-py repo https://github.com/redis/redis-py/pull/2607 
The biggest change from 4.3.5 to 4.4.0 was that they got rid of second buffer `SocketBuffer` and started using `asyncio.StreamReader` directly. Which resulted in that odd behaviour when canceling a task executing redis commands

jajce:
@carltongibson yeah, unfortunately my PR in redis-py is supposed to fix only `test_receive_cancel` :(. I am still looking into what is happening with `test_message_expiry__group_send__one_channel_expires_message`

jajce:
Ok, investigated a little bit more and will share what I've found so far. Now I am pretty much sure that both of the issues are related to the removal of `SocketBuffer` https://github.com/redis/redis-py/pull/2418/ . The `test_message_expiry__group_send__one_channel_expires_message` issue is very well described in this ticket https://github.com/redis/redis-py/issues/2579 . 
In our test the issue is that even though we time out our `receive` here:
https://github.com/django/channels_redis/blob/89b29ad300704aa09ec9e2e31efad5eec11ec03f/tests/test_core.py#L589-L591
our next `receive`, which is here
https://github.com/django/channels_redis/blob/89b29ad300704aa09ec9e2e31efad5eec11ec03f/tests/test_core.py#L594
will have to wait for 5 seconds 
[...]

Issue #319, opened 2022-07-20
v4b1: PubSub receive cleanup hang


bbrowning918:
Following discussion on https://github.com/django/channels_redis/pull/317

On 4.0.0b1, the `test_groups_basic` in either [test_pubsub.py](https://github.com/django/channels_redis/blob/d946ce2875b2a2efe2c4cbfb14107fd7d7a4a551/tests/test_pubsub.py#L90) and [test_pubsub_sentinel.py](https://github.com/django/channels_redis/blob/d946ce2875b2a2efe2c4cbfb14107fd7d7a4a551/tests/test_pubsub_sentinel.py#L78) tests can hang intermittently. This is most pronounced on CI environments (GitHub actions for this repo show some examples for PRs), and locally for myself occurs roughly every 6-8 runs of the below snippet.

The hang occurs with a `RedisPubSubChannelLayer` when checking that a message is not received on some particular channel, this is a small test to more easily produce the issue for `test_pubsub`:
```
@pytest.mark.asyncio
async def test_receive_hang(channel_layer):
    channel_name = await channel_layer.new_channel(prefix="test-channel")
    with pytest.raises(asyncio.TimeoutError):
        async with async_timeout.timeout(1):
            await channel_layer.receive(channel_name)
```

Preliminary tracing found [`receive`](https://github.com/django/channels_redis/blob/main/channels_redis/pubsub.py#L194) on attempting to [`unsubscribe`](https://github.com/django/channels_redis/blob/main/channels_redis/pubsub.py#L290) fails to ever return a connection from `_get_sub_conn`.

A [`_receive_task`](https://github.com/django/channels_redis/blob/d946ce2875b2a2efe2c4cbfb14107fd7d7a4a551/channels_redis/pubsub.py#L356) appears to never return on multiple attempts, holding a lock indefinitely.

The following print annotations,
```
    async def _get_sub_conn(self):
        if self._keepalive_task is None:
            self._keepalive_task = asyncio.ensure_future(self._do_keepalive())
        if self._lock is None:
            self._lock = asyncio.Lock()
        print(self._lock)
        async with self._lock:
            if self._sub_conn is not None and self._sub_conn.connection is None:
                await self._put_redis_conn(self._sub_conn)
                self._sub_conn = None
                self._notify_consumers(self.channel_layer.on_disconnect)
            if self._sub_conn is None:
                if self._receive_task is not None:
                    print(self._receive_task)
                    self._receive_task.cancel()
                    try:
[...]

PR #349, opened 2023-02-28
Pin redis py to 4.3.5

bbrowning918:
This in practice seems to be the last/most stable version of redis-py for the test suite. GitHub's default is to run for 6 hours which is excessive, if tox hangs and takes longer than 10 minutes it never comes back.

Would make a temporary fix to get proper support/fix for #348

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,6 +10,7 @@ jobs:
   tests:
     name: Python ${{ matrix.python-version }}
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     strategy:
       fail-fast: false
       matrix:
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ setup(
     include_package_data=True,
     python_requires=">=3.7",
     install_requires=[
-        "redis>=4.2.0",
+        "redis~=4.3.5",
         "msgpack~=1.0",
         "asgiref>=3.2.10,<4",
         "channels",


bbrowning918:
Stuck again on our old friend `test_receive_hang`.

bbrowning918:
The tests hang consistently in the same spot, but not in the same python version. Sister runs in [my fork](https://github.com/bbrowning918/channels_redis/actions) will fail on one or more different python versions than here, and even more puzzling can succeed from time to rime.

`tests/test_pubsub_sentinel.py::test_recieve_hang` is where the tests get caught up. I am trying my usual prodding to figure out where/why.

carltongibson:
Thanks for your work on this @bbrowning918! 🎁 

It's #319 again/still 😅

`test_receive_hang`… Q: Do you think it's worth to skip this and add to #348, to allow a bit of time to resolve? 🤔

bbrowning918:
I had considered that but am unsure, which is why I moved this to a draft. 

#319 was certainly improved upon and lessened. That an added regression test for that issue is suffering from a hang hints something still lurks. 

Should #348  be resolved given it shows a fail and a hang in a different spot (and I believe earlier in the suite) then skipping right to `redis-py >= 4.4.0` where the suite passes every time would be satisfactory. My goal here is to reduce false positives that block or cloud merging of other's excellent contributions and improvements. 

I do not know what kind of noise me aggressively retrying the actions creates for yourself and other maintainers. I do not want that to be intrusive as I stumble about experimenting with stability. I have no issue closing this and playing off in a fork until (or if) a meaningful result.

carltongibson:
Hey @bbrowning918 — no noise, please do experiment away! 😀

bbrowning918:
Ok, 3 for 3 in **both** fork and main is something. The tests themselves pass, but the pytest fixture for the channel layer hangs on it's call to `await channel_layer.flush()` for pubsub with and without a sentinel.

I am yet unable to consistently reproduce the flush hang  locally and write a test against it.

carltongibson (approved):
OK, let's take this for now. 

Using redis 4.3.5 I'm not seeing the hang locally, whereas as soon as I move to 4.4.x or above I do.

PR #351, opened 2023-03-01
docs: Update README ref to Redis to match package requirement

### LINKED ISSUE #350: Redis dependency discrepancy
I'm in the process of upgrading a project to django 4.2 and all things channels v4 and I've got web sockets failing so I'm going back to basics to verify the project setup.

Going over the readme I can see a requirement of `Redis >= 5.0` (this version changed in 2020 [here](https://github.com/django/channels_redis/commit/5ddd906efdcae55fc5c3a356a0408acb3c63afdd)), but `pip-tools` was happy with me still using `redis==4.5.1` so I dug a little deeper.

In the installation requirements for v4 and the actual requirement is at `redis>=4.2.0` ([see here](https://github.com/django/channels_redis/blob/95d87449c2b899c1cfdf8a10cfa1425e30127ef2/setup.py#L33)). 

I assume therefore that redis 4 is still ok to use, but this may confuse anyone going through this same process as me.

marksweb:
This fixes #350 to update the README redis dependency to match the actual python dependency.

Nice & simple.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -222,7 +222,7 @@ And then in your channels consumer, you can implement the handler:
 Dependencies
 ------------
 
-Redis >= 4.2.0 is required for `channels_redis`. Python 3.7 or higher is required.
+Redis server >= 5.0 is required for `channels_redis`. Python 3.7 or higher is required.
 
 
 Used commands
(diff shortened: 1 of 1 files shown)


marksweb:
@carltongibson ok so using `redis>=5` is more of a piece of advice to avoid issues & therefore no change required?

carltongibson:
I suggested an edit. How does that go for you?

marksweb:
@carltongibson Yeah that seems fair 👍

PR #357, opened 2023-03-28
Updated GHA workflow action versions.

carltongibson:


carltongibson:
See #348 for Python 3.11 failures, which are awaiting the **next** version of `redis-py`.

PR #363, opened 2023-04-08
Use normalized `channels-redis` PyPI package name in README

johnthagen:
Consistently use the package name `channels-redis` throughout the README when referring to the PyPI package name.


carltongibson (approved):
Ok, yep. Great

Issue #330, opened 2022-09-28
Missing documentation on how to set `connection_kwargs` params.


slyapustin:
What is the right way to specify `connection_kwargs`?

Based on the changelog - it was implemented a while ago, but there are no examples on how to set that and no mentions in the docs.

I'm happy to submit PR, but I can't figure out how to set them on my own 🤔 
>
1.3.0 (2017-04-07)
------------------

* Change the format of connection arguments to be a single dict called
  ``connection_kwargs`` rather than individual options, as they change by
  connection type. You will need to change your settings if you have any of
  socket_connect_timeout, socket_timeout, socket_keepalive or
  socket_keepalive_options set to move them into a ``connection_kwargs`` dict.

carltongibson:
`connection_kwargs` will just be passed to the underlying library no?

slyapustin:
@carltongibson How it should look in the `CHANNEL_LAYERS` configuration?

```python
CHANNEL_LAYERS = {
    "default": {
        "BACKEND": "channels_redis.core.RedisChannelLayer",
        # how to pass `socket_connect_timeout`, `socket_timeout` and etc?
    }
}
```

slyapustin:
This might be relevant to #331

carltongibson:
This should be resolved by https://github.com/django/channels_redis/pull/370.

Issue #235, opened 2021-01-20
Feature request: configure SSL settings in CHANNEL_LAYERS


jray-afk:
Hi, I posted about my adventure trying to configure SSL for Redis in the suggested Google Groups page which has all the requested information prepopulated in the body of this issue form, haven't gotten a response yet: https://groups.google.com/g/django-users/c/S094vik_gW0

After digging into the channels_redis code in this repo I realized that I probably need to submit a new feature request here in order to configure SSL settings in CHANNEL_LAYERS. (Unless, do you know of another way to configure SSL using channels_redis?)

Thank you!

carltongibson:
So... we'd need to by able to provide an `SSLContext` when creating connections. 

See [`aioredis.create_connection()`](https://aioredis.readthedocs.io/en/v1.3.0/api_reference.html#aioredis.create_connection). 

Happy to accept input here.

jray-afk:
@carltongibson I appreciate the response. Yes that makes sense.

For now I will set up symmetric_encryption_keys to add some level of protection, but ultimately I would prefer setting up SSL if possible.

carltongibson:
Super! If you get started and need some input, let me know. Thanks! 👍

skadz:
I was just running in to this issue myself today, so happy to help test!

jray-afk:
@skadz glad to hear I'm not the only one! Haven't had a chance to fork off of this project and try to implement the changes, but I may have to go rogue and go off of the version I built my project on, 2.3.2

skadz:
> @skadz glad to hear I'm not the only one! Haven't had a chance to fork off of this project and try to implement the changes, but I may have to go rogue and go off of the version I built my project on, 2.3.2

@jray-afk Does this mean you actually have made the changes and got it working and just need to do the fork/submit PR or that you hacked some sort of other "fix" in?  I'd like to get this implemented ASAP and if you have something that is working, I'd love to help you get it in to the main code, but also can help test anything you've done so far (or would like to know about how you hacked it in ;)

nmacianx:
@jray-afk @skadz any updates here? I'm facing the very same issue and help would be much appreciated!!

carltongibson:
Just awaiting someone to pick it up and submit a pull request I think

dablak:
[...]

PR #370, opened 2023-09-01
Document Redis SSL configs

bbrowning918:
There is ongoing discussion and help in https://github.com/django/channels_redis/issues/235 and https://github.com/django/channels_redis/issues/330 (are there more?) about how to specify extra config for the connections to Redis, particularly SSL related.

The example provided is how we got Heroku to behave with it's self-signed certs on 4.1.0. Hopefully it will be helpful and answer a common question.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -65,9 +65,21 @@ Possible options for ``CONFIG`` are listed below.
 ``hosts``
 ~~~~~~~~~
 
-The server(s) to connect to, as either URIs, ``(host, port)`` tuples, or dicts conforming to `redis Connection <https://redis-py.readthedocs.io/en/v4.3.3/connections.html#redis.connection.Connection>`_.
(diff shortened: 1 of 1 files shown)


bbrowning918:
:disappointed: oh

carltongibson (approved):
Thanks for this @bbrowning918 👍 

(I need to look at  what happened to the CI...)

justuswilhelm:
May I suggest that there could be a typo in the README as it is right now? (not mentioning the missing restructured text code block)

Right now it says something like the following, which is not a valid dict.

```
   "hosts":[
        "address": "rediss://user@host:port",  // "REDIS_TLS_URL"
        "ssl_cert_reqs": None
    ]
```

I reverted to using my own helper for now. It's too bad testing deployments on Heroku is so time consuming.

bbrowning918:
@justuswilhelm In hindsight I did not do a great job 'copy-pasting' our config from memory. Thank you for noticing that and the rst syntax mistake.

carltongibson:
I tweaked it, but happy to take any improvements.

PR #374, opened 2023-11-23
test: Add python 3.12 and drop 3.7

marksweb:
This adds python 3.12 to the tests and drops 3.7


carltongibson (approved):
Look at that. Lovely 🎁

smithdc1:
Should this be increased too? 

https://github.com/django/channels_redis/blob/main/setup.py#L32

marksweb:
> Should this be increased too?
> 
> https://github.com/django/channels_redis/blob/main/setup.py#L32

Good spot David. I can update, though there are reasons why that might not want to be updated (yet).

carltongibson:
Yes, good spot. (Likely there's a reference or two elsewhere, README say?, that a search would catch.)

marksweb:
I'll take a look before I get going with the same for channels

PR #375, opened 2023-11-23
chore: Cleanup refs to python 3.7

marksweb:
This cleans up the python requires and readme that still referred to python 3.7 after I've removed it from the tests.


carltongibson (approved):
Great. Thanks.

PR #376, opened 2024-01-05
Avoid redis 5.x deprecation warning when closing connection

sevdog:
[Redis 5.0.1](https://github.com/redis/redis-py/releases/tag/v5.0.1) adds a new method `aclose` and deprecates the `close` method in the async client.

This change aims to resolve the deprecation warnings which arise when using the new redis-py until redis 4.5 is still supported.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -15,7 +15,13 @@ from redis import asyncio as aioredis
 from channels.exceptions import ChannelFull
 from channels.layers import BaseChannelLayer
 
-from .utils import _consistent_hash, _wrap_close, create_pool, decode_hosts
+from .utils import (
+    _close_redis,
+    _consistent_hash,
+    _wrap_close,
+    create_pool,
+    decode_hosts,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -86,7 +92,7 @@ class RedisLoopLayer:
         async with self._lock:
             for index in list(self._connections):
                 connection = self._connections.pop(index)
-                await connection.close(close_connection_pool=True)
+                await _close_redis(connection)
 
 
 class RedisChannelLayer(BaseChannelLayer):
--- a/channels_redis/pubsub.py
+++ b/channels_redis/pubsub.py
@@ -6,7 +6,13 @@ import uuid
 import msgpack
 from redis import asyncio as aioredis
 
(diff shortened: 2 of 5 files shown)


carltongibson (commented):
Hey @sevdog. Thanks for this. Looks sensible. 

Q: Do we need to add testing for the different versions of redis-py? (Perhaps just for the main tests?)

carltongibson:
I think that might be helpful... — we could then easily test against (e.g.) current and upcoming versions, and perhaps be clearer/stricter about what versions we support. 

What's your take? 

Fancy adding it? 🎁

sevdog:
Taht could be nice!

I think that a redis compatibility-matrix may be set-up in the scope of this PR.

I would also like to add the same testing mechanism also for channels (since currently [no version is pinned in setup](https://github.com/django/channels_redis/blob/66187aa3d96b11bc67a93d5f1cdb2c54f4717bb2/setup.py#L37), there is a generic statement in the changelog but nothing else).

Maybe I could work on a PR which adds `channels` and `redis` to test matrix before merging this one.

carltongibson:
> Maybe I could work on a PR which adds channels and redis to test matrix before merging this one.

That would be perfect 🤩

Not sure we need Channels... but if you think so — Did it ever fail? 🤔

I don't think we need to test every combination of channels against every redis 🤔, and I'd just run the core tests, rather than all the sentinel and cluster ones. (It seems unlikely there's an extra regression there so saving a few cycles is probably better 🌳)

sevdog:
Let me try to propose a test matrix:
|                  | redis develop            | redis stable       | redis 4.5                |
|------------------|--------------------------|--------------------|--------------------------|
| channels develop | :heavy_multiplication_x: | :heavy_check_mark: | :heavy_multiplication_x: |
| channels stable  | :heavy_check_mark:       | :heavy_check_mark: | :heavy_check_mark:       |
| channels 3.0     | :heavy_multiplication_x: | :heavy_check_mark: | :heavy_multiplication_x: |

Writing it down: redis compatibility is tested only against the latest release of channels and channels compatibility is tested only angainst the latest release of redis.

carltongibson:
That looks great

carltongibson (approved):
Great. Let's have it. Thanks @sevdog

PR #380, opened 2024-01-25
support msgpack.packb kwargs

mpnowacki-reef:
The main reason for submitting this PR is to give users the ability to subclass RedisChannelLayer and do sth like: 

```
def default(value):
    if isinstance(value, enum.Enum):
        return value.value
    return value


class ECRedisChannelLayer(RedisChannelLayer):
    """EC stands for "enum-compatible" """
    def _packing_kwargs(self, message):
        return {**super()._packing_kwargs, 'default': default}
```


carltongibson (commented):
Could you add a test case showing the usage here?

PR #388, opened 2024-05-03
Update README.rst

kofnet002:
I followed the tutorial and it was great, but I had the error below
 ```TypeError: RedisChannelLayer.__init__() got an unexpected keyword argument 'ssl_cert_reqs'```.

because the CHANNELS_LAYER CONFIG was like this
```"BACKEND": "channels_redis.core.RedisChannelLayer"```

I had to uninstall the channels redis, ``channels-redis``, and install  ``channels_redis`` for it to work. And use your configuration as below:

```CHANNEL_LAYERS = {   
"default": {
    "BACKEND": "channels_redis.pubsub.RedisPubSubChannelLayer",
    "CONFIG": {
        "hosts":[{
            "address": "rediss://:your_redis_url",  # "REDIS_TLS_URL"
            "ssl_cert_reqs": None,
        }]
    }
}
}

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,7 +28,7 @@ Installation
 
 .. code-block::
 
-    pip install channels-redis
+    pip install channels_redis
 
 **Note:** Prior versions of this package were called ``asgi_redis`` and are
 still available under PyPI as that name if you need them for Channels 1.x projects.


carltongibson (commented):
These are the same package. 

https://pypi.org/project/channels-redis/

Here's the PEP: 

> The name should be lowercased with all runs of the characters ., -, or _ replaced with a single - character. 

https://peps.python.org/pep-0503/#normalized-names

kofnet002:
Interesting but I had to install the package “channel_redis” instead of
“channel-redis” for it to work. I stand to be corrected.
Thank you

Issue #222, opened 2020-09-04
Redis UTF-8 cause hard to debug issue on group_send


insspb:
All code is reproducible in multichat example app with last django and channels_redis.

It is possible to make `settings.py` configuration like this:
```
CHANNEL_LAYERS = {
    "default": {
        # This example app uses the Redis channel layer implementation channels_redis
        "BACKEND": "channels_redis.core.RedisChannelLayer",
        "CONFIG": {
            "hosts": [
                "redis://redis:6379/0?encoding=utf-8"
            ]
        },
    },
}
```

Let's take code that will create fake group and send message there.
Code example for pycharm django console.
```
from asgiref.sync import async_to_sync
from channels.layers import get_channel_layer

layer = get_channel_layer()
add = async_to_sync(layer.group_add)
send = async_to_sync(layer.group_send)
res1 = add("some", "random")
res2 = send("some", {"message": "message",})
```

You will get this traceback: 
```
Traceback (most recent call last):
  File "<input>", line 8, in <module>
  File "/home/insspb/.venvs/quick/lib/python3.7/site-packages/asgiref/sync.py", line 139, in __call__
    return call_result.result()
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 428, in result
    return self.__get_result()
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/home/insspb/.venvs/quick/lib/python3.7/site-packages/asgiref/sync.py", line 204, in main_wrap
    result = await self.awaitable(*args, **kwargs)
  File "/home/insspb/.venvs/quick/lib/python3.7/site-packages/channels_redis/core.py", line 634, in group_send
    x.decode("utf8") for x in await connection.zrange(key, 0, -1)
  File "/home/insspb/.venvs/quick/lib/python3.7/site-packages/channels_redis/core.py", line 634, in <listcomp>
    x.decode("utf8") for x in await connection.zrange(key, 0, -1)
AttributeError: 'str' object has no attribute 'decode'
```

And it is very hard to guess that problem in connection string. Because item in redis will be created.

If replace connection code with same, but without encoding mention all will work without any issues.

```
CHANNEL_LAYERS = {
    "default": {
        # This example app uses the Redis channel layer implementation channels_redis
        "BACKEND": "channels_redis.core.RedisChannelLayer",
        "CONFIG": {
            "hosts": [
                "redis://redis:6379/0"
            ]
        },
    },
}
```
[...]

PR #398, opened 2024-08-21
Add serialization registry

sevdog:
This PR aims at two goals:

1. Provide a common serialization/deserialization layer between `RedisChannelLayer` and `PubSubRedisChannelLayer`
2. Provide a way to customize how messages are serialized before being forwarded to the broker (aka: redis), which may address problems like https://github.com/django/channels/issues/2094

The serializer registry is somehow inspired by that of `django.core.serializers` package.
This also prepares a future in which msgpack would no longer be a mandatory dependency.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -171,6 +171,7 @@ to 10, and all ``websocket.send!`` channels to 20:
 If you want to enforce a matching order, use an ``OrderedDict`` as the
 argument; channels will then be matched in the order the dict provides them.
 
+.. _encryption
 ``symmetric_encryption_keys``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -236,6 +237,51 @@ And then in your channels consumer, you can implement the handler:
 
     async def redis_disconnect(self, *args):
         # Handle disconnect
+
+
+
+``serializer_format``
+~~~~~~~~~~~~~~~~~~~~~~
+By default every message sent to redis is encoded using `msgpack <https://msgpack.org/>`_ (_currently ``msgpack`` is a mandatory dependency of this package, it may become optional in a future release_).
+It is also possible to switch to `JSON <http://www.json.org/>`_:
+
+.. code-block:: python
+
+    CHANNEL_LAYERS = {
+        "default": {
+            "BACKEND": "channels_redis.core.RedisChannelLayer",
+            "CONFIG": {
+                "hosts": ["redis://:password@127.0.0.1:6379/0"],
+                "serializer_format": "json",
+            },
+        },
+    }
+
+
+Custom serializer can be defined by:
+
+- extending ``channels_redis.serializers.BaseMessageSerializer``, implementing ``as_bytes `` and ``from_bytes`` methods
+- using any class which accepts generic keyword arguments and provides ``serialize``/``deserialize`` methods
+
+Then it may be registered (or can be overriden) by using ``channels_redis.serializers.registry``:
+
+.. code-block:: python
+
+    from channels_redis.serializers import registry
+
+    class MyFormatSerializer:
+        def serialize(self, message):
+            ...
+        def deserialize(self, message):
+            ...
+
+    registry.register_serializer('myformat', MyFormatSerializer)
+
+**NOTE**: the registry allows to override the serializer class used for a specific format without any particular check nor constraint, thus it is recommended to pay attention with order-of-imports when using third-party serializers which may override a built-in format.
+
+
+Serializers are also responsible for encryption *symmetric_encryption_keys*. When extending  ``channels_redis.serializers.BaseMessageSerializer`` encryption is already configured in the base class, unless you override ``serialize``/``deserialize`` methods: in this case you should call ``self.crypter.encrypt`` in serialization and ``self.crypter.decrypt`` in deserialization process. When using full custom serializer expect an optional sequence of keys to be passed via ``symmetric_encryption_keys``.
+
 
 Dependencies
 ------------
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -1,20 +1,17 @@
 import asyncio
-import base64
 import collections
 import functools
-import hashlib
 import itertools
 import logging
-import random
 import time
 import uuid
 
-import msgpack
 from redis import asyncio as aioredis
 
 from channels.exceptions import ChannelFull
 from channels.layers import BaseChannelLayer
 
+from .serializers import registry
 from .utils import (
     _close_redis,
     _consistent_hash,
@@ -115,6 +112,8 @@ class RedisChannelLayer(BaseChannelLayer):
         capacity=100,
         channel_capacity=None,
         symmetric_encryption_keys=None,
+        random_prefix_length=12,
+        serializer_format="msgpack",
     ):
         # Store basic information
         self.expiry = expiry
@@ -126,6 +125,14 @@ class RedisChannelLayer(BaseChannelLayer):
         # Configure the host objects
         self.hosts = decode_hosts(hosts)
         self.ring_size = len(self.hosts)
+        # serialization
+        self._serializer = registry.get_serializer(
+            serializer_format,
+            # As we use an sorted set to expire messages we need to guarantee uniqueness, with 12 bytes.
+            random_prefix_length=random_prefix_length,
+            expiry=self.expiry,
+            symmetric_encryption_keys=symmetric_encryption_keys,
+        )
         # Cached redis connection pools and the event loop they are from
         self._layers = {}
         # Normal channels choose a host index by cycling through the available hosts
@@ -133,8 +140,6 @@ class RedisChannelLayer(BaseChannelLayer):
         self._send_index_generator = itertools.cycle(range(len(self.hosts)))
         # Decide on a unique client prefix to use in ! sections
         self.client_prefix = uuid.uuid4().hex
-        # Set up any encryption objects
-        self._setup_encryption(symmetric_encryption_keys)
         # Number of coroutines trying to receive right now
         self.receive_count = 0
         # The receive lock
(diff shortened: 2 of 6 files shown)


carltongibson:
Oh, very exciting @sevdog — Thanks for thinking about this.

sevdog:
I would like to know your opinion @carltongibson regarding my design choice to put encryption into the serialization process.

Also I noted that currently thare are no tests for that feature, I could add them to this PR if you consider that useful.

carltongibson:
@sevdog Let me have a look/think. I've asked @bigfootjon and @acu192 to have a look too.

channels_redis/serializers.py
```diff
@@ -0,0 +1,140 @@
...
+        assert isinstance(serializer_class, type) and (
+            issubclass(serializer_class, BaseMessageSerializer)
+            or hasattr(serializer_class, "serialize")
+            and hasattr(serializer_class, "deserialize")
+        ), """
+            `serializer_class` should be a class which implements `serialize` and `deserialize` method
+            or a subclass of `channels_redis.serializers.BaseMessageSerializer`
+        """
+
+        self._registry[format] = serializer_class
```
bigfootjon, line 127: Should this throw an exception if the `format` is already registered with a different type?

(IMO we should allow users to call `register_serializer` multiple times with the same `serializer_class` to prevent `import` ordering messing things up)

tests/test_serializers.py
```diff
@@ -0,0 +1,48 @@
+import pytest
```
bigfootjon, line 1: I think we should add test coverage for `JSONSerializer` here.

And also msgpack (though I assume that already has some coverage, doesn't hurt to add more)

README.rst
```diff
@@ -237,6 +238,44 @@ And then in your channels consumer, you can implement the handler:
...
+A new serializer may be registered (or can be overriden) by using ``channels_redis.serializers.registry``,
+providing a class which extends ``channels_redis.serializers.BaseMessageSerializer``, implementing ``dumps``
+and ``loads`` methods, or which provides ``serialize``/``deserialize`` methods and calling the registration method on registry:
+
+.. code-block:: python
+
+    from channels_redis.serializers import registry
+
+    class MyFormatSerializer:
+        def serialize(self, message):
```
bigfootjon, line 269: Can you update the docs to match the actual code? Per the code below the user needs to implement `dumps` and `loads` (see my comment there though)

Additionally, the docs should clarify what type the inputs (i.e. `message`) and output (the `return` value) have.

(e.g. `dumps` should return `bytes`)

channels_redis/serializers.py
```diff
@@ -0,0 +1,140 @@
...
+        raise NotImplementedError
+
+    def serialize(self, message):
+        """
+        Serializes message to a byte string.
+        """
+        message = self.dumps(message)
+        # ensure message is bytes
+        if isinstance(message, str):
+            message = message.encode("utf-8")
```
bigfootjon, line 55: I'm not sure I like this specific part of the design. IMO the serializer implementation should handle *all* serialization, including the string encoding settings (`utf-8` or otherwise)

Deferring all serialization to the `dumps` function means that this is symmetric with how `deserialize` calls `load`

(this raises a separate concern that `dumps` and `loads` are not named correctly, perhaps `as_bytes` and `from_bytes` instead?)

channels_redis/serializers.py
```diff
@@ -0,0 +1,140 @@
...
+            message = self.crypter.encrypt(message)
+
+        if self.random_prefix_length > 0:
+            # provide random prefix
+            message = (
+                random.getrandbits(8 * self.random_prefix_length).to_bytes(
+                    self.random_prefix_length, "big"
+                )
+                + message
+            )
```
bigfootjon, line 66: Can you add some tests covering this logic? Making `random_prefix_length` dynamic seems reasonable to me but we need coverage.

channels_redis/core.py
```diff
@@ -650,31 +640,6 @@ def _group_key(self, group):
...
-        """
-        Deserializes from a byte string.
-        """
-        # Removes the random prefix
-        message = message[12:]
-
-        if self.crypter:
-            message = self.crypter.decrypt(message, self.expiry + 10)
-        return msgpack.unpackb(message, raw=False)
-
```
bigfootjon, line 677: Is there a chance people were calling these functions directly?

bigfootjon (changes requested):
I love the idea! I have a few questions and some requests but this is in great shape already!

channels_redis/pubsub.py
```diff
@@ -44,18 +57,6 @@ def __getattr__(self, name):
...
-        """
-        Serializes message to a byte string.
-        """
-        return msgpack.packb(message)
-
-    def deserialize(self, message):
-        """
-        Deserializes from a byte string.
-        """
-        return msgpack.unpackb(message)
```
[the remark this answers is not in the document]
sevdog: If they were overridden and called with `super()` this may break.

I can keep these methods for backward compatibility and just remove the "assignment" in the `__init__` method.

channels_redis/serializers.py
sevdog: Seems reasonable, I used `utf-8` as it is also used in [`group_send`](https://github.com/django/channels_redis/blob/13cef4502a95fbb7f41bb97eaa6f4d02c1f7c91c/channels_redis/core.py#L533) and in [`_group_key`](https://github.com/django/channels_redis/blob/13cef4502a95fbb7f41bb97eaa6f4d02c1f7c91c/channels_redis/core.py#L651).

Thus I belive that if we provide support for encoding for messages we may also extend that to those methods to keep consistency (encoding also relates to #222).

channels_redis/serializers.py
```diff
@@ -0,0 +1,140 @@
...
+    def __init__(self, *args, **kwargs):
+        raise self.exception
+
+
+class JSONSerializer(BaseMessageSerializer):
+    dumps = staticmethod(json.dumps)
+    loads = staticmethod(json.loads)
+
+
+# code ready for a future in which msgpack may become an optional dependency
```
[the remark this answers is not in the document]
sevdog: Currently I think that there should be an announce to inform people know that msgpack an opt-in dependency, to avoid issue in projects which only have `channels-redis` as dependency. Something like:

> In version X.X.X msgpack will become an optional dependency and thus you will need to install it by your own.

Also We can also add a `warning` whenever a `MissinSerializer` is istantiated (thus will result in a warning + exception during tests).

channels_redis/serializers.py
```diff
@@ -0,0 +1,140 @@
...
+        self._registry = {}
+
+    def register_serializer(self, format, serializer_class):
+        """
+        Register a new serializer for given format
+        """
+        assert isinstance(serializer_class, type) and (
+            issubclass(serializer_class, BaseMessageSerializer)
+            or hasattr(serializer_class, "serialize")
+            and hasattr(serializer_class, "deserialize")
```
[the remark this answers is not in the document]
sevdog: Ok, I checked that black does not complain for this thus I can add it.

In my mind they were "useless" since I am very used to read boolean expressions and to know the precedence of operators. :smile:

channels_redis/serializers.py
sevdog: Currently I left possible this to override an existing serializer. It could be done via an extra parameter but it would not resolve the problem of import/loading problem.

Also the serializer registry in django core does not raise any error nor performs any check when a serializer is re-registered
https://github.com/django/django/blob/fd1dd767783b5a7ec1a594fcc5885e7e4178dd26/django/core/serializers/__init__.py#L85-L88

channels_redis/serializers.py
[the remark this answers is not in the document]
bigfootjon: Makes sense, I agree that should be a later change

channels_redis/serializers.py
bigfootjon: Coolio, I'm fine with this as-is then

channels_redis/serializers.py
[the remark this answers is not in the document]
carltongibson: > ... know the precedence of operators...

Shouldn't have people need to stop and think here. +1 to the extra parentheses. 


Although... 🤔 Could we check `isinstance` on a `Protocol` here? (Perhaps not but...)

carltongibson (commented):
This all looks quite reasonable. 

I'm not quite seeing how it ties into the memory leak issue... 🤔 (Do we have an actual minimal reproduce for that anywhere?)

sevdog:
I belive the usage of `Protocol` could be added later, when a more extensive support for _typing_ gets into the codebase.

carltongibson:
OK thanks @sevdog — let me have a look at that. I have no objection to this proposal. 👍

README.rst
```diff
@@ -237,6 +238,51 @@ And then in your channels consumer, you can implement the handler:
...
+        },
+    }
+
+
+Custom serializer can be defined by:
+
+- extending ``channels_redis.serializers.BaseMessageSerializer``, implementing ``dumps`` and ``loads`` methods
+- using any class which accepts generic keyword arguments and provides ``serialize``/``deserialize`` methods
+
+Then it may be registerd (or can be overriden) by using ``channels_redis.serializers.registry``:
```
bigfootjon, line 266: Typo: registerd -> registered

README.rst
```diff
@@ -237,6 +238,51 @@ And then in your channels consumer, you can implement the handler:
...
+                "hosts": ["redis://:password@127.0.0.1:6379/0"],
+                "serializer_format": "json",
+            },
+        },
+    }
+
+
+Custom serializer can be defined by:
+
+- extending ``channels_redis.serializers.BaseMessageSerializer``, implementing ``dumps`` and ``loads`` methods
```
bigfootjon, line 263: I think this needs to be updated to `as_bytes` and `from_bytes` right?

channels_redis/pubsub.py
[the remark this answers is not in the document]
bigfootjon: Yeah my preference is that we restore this, just for backwards compat reasons

channels_redis/serializers.py
```diff
@@ -0,0 +1,170 @@
...
+    def get_serializer(self, format, *args, **kwargs):
+        try:
+            serializer_class = self._registry[format]
+        except KeyError:
+            raise SerializerDoesNotExist(format)
+
+        return serializer_class(*args, **kwargs)
+
+
+# register default serializers
```
bigfootjon, line 167: I'm not sure this comment adds any value, can you remove?

bigfootjon (changes requested):
Great work! Sorry it took me a bit to get back to this and review.

Just a few comments left, mostly just cleanup at this point

README.rst
sevdog: You are right, I forgot to update the docs!

sevdog:
Changes pushed.

bigfootjon (approved):
LGTM! Thanks for working through all my comments @sevdog

@carltongibson, when you get a chance could you do a pass on this PR and merge if you agree?

carltongibson (approved):
This looks good. Sorry for the slow follow up, I missed your last ping @bigfootjon.

I just want to make a couple of tweaks to the README changes, which I'll do when I'm at the computer and than I'll pull this in. 

Thanks @sevdog. Really nice addition.

carltongibson:
Will merge via #401.

Issue #399, opened 2024-09-14
Allow sending messages in bulk


olzhasar:
# Feature request

Sometimes there is a need to send multiple messages to a channel (or group) at once. While this can be done using a group of `asyncio` tasks, the underlying logic will be suboptimal, cause there would be multiple redis calls for discarding old messages, checking the capacity, appending to the set and setting the expiration.

It would be cool if there were optimized functions for that using a single `zadd` call.
Something like this:

```
    async def send_multiple(self, channel, messages: list):
        ...

    async def group_send_multiple(self, channel, messages: list):
        ...
``` 

I am not sure if such an issue should be created in the `channels` repo first, let me know if that's the case.

carltongibson:
This isn't a bad idea. We'd probably need to add something to the base channel layer, and docs too. 

Do you want to experiment with an implementation?

olzhasar:
Sure, I will prepare a PR for this once I have some free time.

PR #400, opened 2024-10-03
Allow sending multiple messages in bulk

olzhasar:
This is a draft implementation for #399

It adds support for sending multiple messages to the core layer only. It probably does not make much sense to add that to the pubsub layer, cause the underlying redis [publish command](https://redis.io/docs/latest/commands/publish/) only supports a single message.

I had to add an additional argument to the lua script for group sending, I hope that's not a big deal. But let me know if it is, I guess we can make the script dynamic.


channels_redis/core.py
```diff
@@ -207,13 +223,14 @@ async def send(self, channel, message):
...
-            channel
-        ):
+        current_length = await connection.zcount(channel_key, "-inf", "+inf")
+
+        if current_length + len(messages) > self.get_capacity(channel):
             raise ChannelFull()
 
         # Push onto the list then set it to expire in case it's not consumed
-        await connection.zadd(channel_key, {self.serialize(message): time.time()})
+        now = time.time()
```
bigfootjon, line 232: Is this call to `time.time()` needed?

channels_redis/core.py
```diff
@@ -521,6 +538,13 @@ async def group_send(self, group, message):
         """
         Sends a message to the entire group.
         """
+        await self.group_send_multiple(group, (message,))
+
+    async def group_send_multiple(self, group, messages):
```
bigfootjon, line 543: Same comment as above, does this need to be a separate function?

bigfootjon (changes requested):
This all looks reasonable to me. I'm not very familiar with Lua, but as long as the test suite covers this and it works I think I'm ok with it

channels_redis/core.py
```diff
@@ -207,15 +220,23 @@ async def send(self, channel, message):
...
         # Push onto the list then set it to expire in case it's not consumed
-        await connection.zadd(channel_key, {self.serialize(message): time.time()})
+        await connection.zadd(channel_key, mapping)
         await connection.expire(channel_key, int(self.expiry))
 
+    def _parse_messages(self, message):
+        """
+        Convert a passed message arg to a tuple of messages.
+        """
+        if not isinstance(message, dict) and hasattr(message, "__iter__"):
```
bigfootjon, line 236: Should we use `isinstance(message, collections.abc.Iterable` here instead? I think that would make this a bit more readable

bigfootjon (changes requested):
@carltongibson: when you have a moment could you take a look at this PR? I think it's ready for you

channels_redis/core.py
carltongibson: I can iterate a dict no?

channels_redis/core.py
```diff
@@ -207,15 +220,23 @@ async def send(self, channel, message):
...
+            return tuple(message)
+        return (message,)
```
carltongibson, line 238: Seems a shame to consume the iterator only to loop over it again in the `for message in messages` loop. 🤔

carltongibson (commented):
I have two queries. 

1. Are we sure about changing the signature of the existing send, rather than adding a new method here? Like, we don't use type hints yet, but think through the change there... from single message type to a union of that and an iterable of such... — does that look like a change we'd make? 
2. Do we need to update the docs and spec for channel layers here? https://channels.readthedocs.io/en/latest/channel_layer_spec.html — I'd think we do. 

Ref 1, I'm slightly wary of adjusting the existing method, as we've had regressions in the Lua scripting before, which are hard to anticipate outside of production.

channels_redis/core.py
olzhasar: Indeed, python `dict` is an instance of `collections.abc.Iterable`

channels_redis/core.py
olzhasar: Good point, will rewrite that

olzhasar:
@carltongibson 

> Are we sure about changing the signature of the existing send, rather than adding a new method here? Like, we don't use type hints yet, but think through the change there... from single message type to a union of that and an iterable of such... — does that look like a change we'd make?

I initially thought the same way and used two separate methods, but [@bigfootjon's comment](https://github.com/django/channels_redis/pull/400#discussion_r1791014902) made me change my mind on that. I agreed cause I thought that typing is out of scope for everything django-related 😁

> I'm slightly wary of adjusting the existing method, as we've had regressions in the Lua scripting before
This is a legit concern. I'm actually okay with closing this PR if risks outweigh possible benefits from this feature. For my own use case, I've already made a workaround in place :)

bigfootjon:
I think the benefits are there so I'd like to see this landed, but if @carltongibson would rather be conservative here I'll defer to his wisdom. @olzhasar would you mind undoing my recommendation and split up each method? Additionally, when you split things up please split the Lua script as well to make sure we don't regress

carltongibson:
("Wisdom" here means only "been bitten by more innocent looking things enough times to be wary" 😅)

olzhasar:
> I think the benefits are there so I'd like to see this landed, but if @carltongibson would rather be conservative here I'll defer to his wisdom. @olzhasar would you mind undoing my recommendation and split up each method? Additionally, when you split things up please split the Lua script as well to make sure we don't regress

I've updated the PR splitting the methods as discussed, and preserving the old lua script for `group_send`.
Tried to reuse as much code as I could in the group sending methods. It's still not ideal, but going any further would likely lead to a complete redesign, which arguably may not be worth it for this particular feature.

bigfootjon (approved):
LGTM, I'll let @carltongibson give it one final lookover

amirreza8002:
hi
[is this](https://github.com/django/channels_redis/pull/400/files#diff-00ee91764606660588fb10cdc9c1e64d5afe345a104ea5ac5b66e63e3d2d59efR205) supposed to be `messages`?

olzhasar:
> hi [is this](https://github.com/django/channels_redis/pull/400/files#diff-00ee91764606660588fb10cdc9c1e64d5afe345a104ea5ac5b66e63e3d2d59efR205) supposed to be `messages`? or `message`?
> 
> and [this](https://github.com/django/channels_redis/pull/400/files#diff-00ee91764606660588fb10cdc9c1e64d5afe345a104ea5ac5b66e63e3d2d59efR553) supposed to be `args`? or `*args`?

Hey,

The first one is a great catch, fixed that.

As for the second thing, it generally can work either way, but I believe the current version leaves more room for future signature changes, e.g. introducing additional positional arguments.

carltongibson:
Hi @olzhasar — sorry for the slow follow up here 🤹 Just to let you know, I'm working currently on some issues on asgiref, and an update for Daphne. I'll get back to channels-redis after that, and this PR is on my list. Thanks for your patience. 🎁

PR #405, opened 2024-12-21
log: use environment to control the log level

ramwin:
In my scenerio, the websocket message must be obtained the client.  
I want to add the error log if the the channel is over capacity so that I can determine whether there were some performance problem.

### DIFF
--- a/channels_redis/core.py
+++ b/channels_redis/core.py
@@ -3,6 +3,7 @@ import collections
 import functools
 import itertools
 import logging
+import os
 import time
 import uuid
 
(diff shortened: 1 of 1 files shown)


ramwin:
Could anyone review this PR? @bigfootjon

bigfootjon:
Hey there, sorry for the slow reply as I was on holiday.

Why is this necessary? Can't you use logging [Handlers](https://docs.python.org/3/howto/logging.html#handler-basic) to route the logging event yourself and handle as needed? I'm confused why changing the log level will do anything to help you achieve your goal

PR #410, opened 2025-03-30
Renamed internal methods in core channel layer.

carltongibson:
Internal API was renamed in Channels 4.2.1

Initial rename will work for new versions, but not older channels versions. Maybe just need to make channels>=4.2.1. a requirement.


carltongibson:
Will add compatibility fallbacks in https://github.com/django/channels/pull/2145 — we can resolve this more leisurely.

shcheklein:
@andrewgodwin @carltongibson what is the status of this, folks? can it be merged / released by chance? Can we make `channels>=4.2.1` in the new release?

carltongibson:
@shcheklein Channels has fallbacks here. Are you hitting an error?

shcheklein:
@carltongibson no, no errors. Tons of warnings on our CI.

carltongibson:
@shcheklein It turns out I got a moment. Here you go: https://pypi.org/project/channels-redis/4.3.0/

PR #417, opened 2025-05-29
FEAT: type hint

iwakitakuma33:


channels_redis/core.py
```diff
@@ -19,6 +19,13 @@
     create_pool,
     decode_hosts,
 )
+from typing import TYPE_CHECKING, Dict, List, Tuple, Union, Optional
+
+if TYPE_CHECKING:
+    from redis.asyncio.connection import ConnectionPool
+    from redis.asyncio.client import Redis
```
iwakitakuma33, line 26: Shouldn't I also use Redis and ConnectionPool?

amirreza8002: those are just classes

what i mean is, if you see something like `Redis().some_method()`, there is a change the type hint on `some_method` is wrong, and we shouldn't just put the same thing in this codebase

channels_redis/serializers.py
```diff
@@ -108,37 +121,43 @@ class JSONSerializer(BaseMessageSerializer):
...
 
+    MsgPackSerializer = _MsgPackSerializer
 else:
 
-    class MsgPackSerializer(BaseMessageSerializer):
+    class __MsgPackSerializer(BaseMessageSerializer):
         as_bytes = staticmethod(msgpack.packb)
         from_bytes = staticmethod(msgpack.unpackb)
 
+    MsgPackSerializer = __MsgPackSerializer
```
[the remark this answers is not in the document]
iwakitakuma33: I intentionally used a different name for avoiding PylancereportRedeclaration.

channels_redis/core.py
```diff
@@ -573,7 +588,12 @@ async def group_send(self, group, message):
             channels_over_capacity = await connection.eval(
                 group_send_lua, len(channel_redis_keys), *channel_redis_keys, *args
             )
-            if channels_over_capacity > 0:
+            _channels_over_capacity = -1
+            try:
+                _channels_over_capacity = float(channels_over_capacity)
+            except Exception:
+                pass
+            if _channels_over_capacity > 0:
```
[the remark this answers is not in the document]
iwakitakuma33: connection.eval is inferred to return await[str] | str.
By converting it to float, it can be compared with 0.

amirreza8002:
hi @bigfootjon 
sorry for the ping
since django is now python >= 3.10, is there any plans to move this package to >=3.10 as well?
if so we could drop a lot of old syntax and take advantage of the newer typing supports

bigfootjon:
Sorry for the slow reply, 2 major pieces of feedback:

1. We need some kind of CI on these changes. Something like how asgiref does it: https://github.com/django/asgiref/blob/main/tox.ini#L11
2. This PR should ONLY add type-hints, please revert all other changes like the `xfail` removals

iwakitakuma33:
@amirreza8002 @bigfootjon 
Thanks for replying.

The xfail fix has been reverted.
The import order has been corrected.
mypy ci has been introduced.
Any has been used where necessary to reduce the number of error messages in mypy results.

iwakitakuma33:
> Sorry for the slow reply, 2 major pieces of feedback:
> 
> 1. We need some kind of CI on these changes. Something like how asgiref does it: https://github.com/django/asgiref/blob/main/tox.ini#L11
> 2. This PR should ONLY add type-hints, please revert all other changes like the `xfail` removals

@amirreza8002 @bigfootjon 
Thanks for replying.

The xfail fix has been reverted.
The import order has been corrected.
mypy ci has been introduced.
Any has been used where necessary to reduce the number of error messages in mypy results.

iwakitakuma33:
> Sorry for the slow reply, 2 major pieces of feedback:
> 
> 1. We need some kind of CI on these changes. Something like how asgiref does it: https://github.com/django/asgiref/blob/main/tox.ini#L11
> 2. This PR should ONLY add type-hints, please revert all other changes like the `xfail` removals

@amirreza8002 @bigfootjon 
Thanks for replying.

The xfail fix has been reverted.
The import order has been corrected.
mypy ci has been introduced.
Any has been used where necessary to reduce the number of error messages in mypy results.

iwakitakuma33:
@amirreza8002 @bigfootjon
I fixed fail of CI.

Could you run workflow?

iwakitakuma33:
@amirreza8002 @bigfootjon
I fixed fail of CI.

Could you approve workflow awaiting?

iwakitakuma33:
@carltongibson 
thanks for approve workflow.

All tests were successful, but only the package dependency resolution error failed.
I have resolved the conflict with the main branch. Can you approve the side workflow?

tox.ini
amirreza8002, line 1: why commit this?
this already exists in the code base

iwakitakuma33: @amirreza8002 
https://github.com/django/channels_redis/pull/417#issuecomment-2954288695

This is because it is necessary to check the type using ci based on the review from bigfootjon.

iwakitakuma33: @amirreza8002 Thank you for your confirmation!

PR #421, opened 2025-12-08
add support for python 3.14

amirreza-sf80:


amirreza-sf80:
hi 👋
let me know if i should drop 3.9
also if i should add classifiers declaring it for free threaded python

.github/workflows/tests.yml
```diff
@@ -20,6 +20,8 @@ jobs:
           - "3.11"
           - "3.12"
           - "3.13"
+          - "3.14"
+          - "3.14t"
```
carltongibson: Do we really need two runs for 3.14 here? 

Are we doing anything with free threading? Is that even a supported configuration for Django currently?

carltongibson (commented):
Hi @amirreza8002. Thanks for this. 

Yes. Classifiers are needed too. 

If you'd like to drop Python 3.9 in a separate commit, that's welcome.

.github/workflows/tests.yml
amirreza-sf80: hi 👋 
I've seen some projects add it, so thought it's a common practice.

I'll drop it then

amirreza-sf80:
think this fixed it

nikolas:
Looks good - I think this can be merged.

Issue #422, opened 2026-06-02
Explicit socket_timeout needed with redis 8


GetPastTheMonkey:
Since updating the `redis` dependency from version `7.4.0` to version `8.0.0`, I get the following exceptions:

```
WebSocket HANDSHAKING /ws/progress-reporting [127.0.0.1:55380]
INFO 2026-06-02 13:16:35,682 django.channels.server WebSocket HANDSHAKING /ws/progress-reporting [127.0.0.1:55380]
WebSocket CONNECT /ws/progress-reporting [127.0.0.1:55380]
INFO 2026-06-02 13:16:35,717 django.channels.server WebSocket CONNECT /ws/progress-reporting [127.0.0.1:55380]
ERROR 2026-06-02 13:16:41,633 daphne.server Exception inside application: Timeout reading from localhost:6379
Traceback (most recent call last):
  File "project/venv/lib/python3.12/site-packages/redis/asyncio/connection.py", line 755, in read_response
    response = await self._parser.read_response(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "project/venv/lib/python3.12/site-packages/redis/_parsers/resp3.py", line 185, in read_response
    response = await self._read_response(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "project/venv/lib/python3.12/site-packages/redis/_parsers/resp3.py", line 197, in _read_response
    raw = await self._readline()
          ^^^^^^^^^^^^^^^^^^^^^^
  File "project/venv/lib/python3.12/site-packages/redis/_parsers/base.py", line 578, in _readline
    data = await self._stream.readline()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/streams.py", line 568, in readline
    line = await self.readuntil(sep)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/streams.py", line 660, in readuntil
    await self._wait_for_data('readuntil')
  File "/usr/lib/python3.12/asyncio/streams.py", line 545, in _wait_for_data
    await self._waiter
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "project/venv/lib/python3.12/site-packages/redis/asyncio/connection.py", line 754, in read_response
    async with async_timeout(read_timeout):
[...]

PR #423, opened 2026-06-08
Fix Redis socket timeout default

camilonova:
Fixes issue #422


camilonova:
I got a lint error on files I didn't change. That's why I had to https://github.com/django/channels_redis/pull/423/commits/ed0e5dad8c3f14fdb55538235bff54013ef9a7e3

PR #424, opened 2026-06-08
Update redis version for tests

camilonova:
As suggested in https://github.com/django/channels_redis/issues/422#issuecomment-4620977065


camilonova:
The lint issue was addressed in https://github.com/django/channels_redis/pull/423

